PR #6, opened 2018-06-08
This contains a new NDT server written in Go.

pboothe:
It only speaks websockets over TLS, and it does not record any web100 info.  It is, however, very fast and very very small: 500 lines in one file instead of 25,000+ lines across 74 files.  It also supports prometheus monitoring for high-quality real-time assessment of server load.

It also eliminates the need for fakewww.

This is an initial import. Almost everything is a copy (from either ndt-server-go or from instead of an edited file.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-cloud/6)


stephen-soltesz:
Because the authorship of this PR makes it ambiguous, I am adding a link to the original history of the changes in ndt-server-go repo that resulted in this PR https://github.com/m-lab/ndt-server-go/commits/onebox/ndt-server.go

PR #20, opened 2018-08-10
ndt-server: add support for TCP BBR

bassosimone:
Based on join work with @stephen-soltesz. While there gitignore
the certificate files that are generated for testing.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-cloud/20)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+/cert.pem
+/key.pem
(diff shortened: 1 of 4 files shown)


bassosimone:
I 100% agree with using port 443 in the future as proposed by @stephen-soltesz. We are going to merge this now to reduce the entropy on our side after a morning in which we performed many experiments. The port-443 change will go in as a subsequent PR.

PR #33, opened 2018-10-17
Improvements after testing on 2G/3G/4G networks

bassosimone:
I've run some more tests in 2G/3G/4G networks. This led me to the following improvements:

1. using a worker for the web client, so the UI is not locked

2. modify the protocol to allow for sending measurement messages with padding, so that the client receives more measurements regardless of the delay caused by 2G/3G/4G buffers

3. other, smaller, miscellaneous, useful fixes

I have also experimented with `TCP_NOTSENT_LOWAT` but that diff is not ready to be included in this PR yet, and I feel like the code in here has been delayed for too much time already. I'll probably ask about this option to @mattmathis later today or in the next days.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-cloud/33)

### DIFF
--- a/html/libndt7.js
+++ b/html/libndt7.js
@@ -12,7 +12,8 @@ const libndt7 = (function () {
     open: 'ndt7.open',
 
     // close is the event emitted when the socket is closed. The
-    // object bound to this event is always null.
+    // object bound to this event is always null. The code SHOULD
+    // always emit this event at the end of the test.
     close: 'ndt7.close',
 
     // error is the event emitted when the socket is closed. The
--- a/html/libndt7.js
+++ b/html/libndt7.js
@@ -26,9 +26,11 @@ const libndt7 = (function () {
     downloadClient: 'ndt7.download.client'
   }
 
+  const version = 0.4
+
   return {
     // version is the client library version.
-    version: 0.3,
+    version: version,
 
     // events exports the events table.
     events: events,
@@ -52,7 +54,26 @@ const libndt7 = (function () {
         if (settings.port) {
           url += ':' + settings.port
         }
-        return url + '/ndt/v7/download'
+        url += '/ndt/v7/download?'
+        for (let key in settings.meta) {
+          if (settings.meta.hasOwnProperty(key)) {
+            const re = /[0-9A-Za-z._]+/
+            if (!key.match(re)) {
+              throw 'Key does not match the expected regular expression: ' + key
+            }
+            let value = settings.meta[key]
+            if (typeof value === 'number' || typeof value === 'boolean') {
+              value += ''  // force conversion to string
+            }
+            if (typeof value !== 'string' || !value.match(re)) {
+              throw 'Value is not a string or does not match the expected ' +
+                    'regular expression: ' + value
+            }
+            url += key + '=' + value + '&'
+          }
+        }
+        url += 'library.name=libndt7.js&library.version=' + version
+        return url
       }
 
       // setupconn creates the WebSocket connection and initializes all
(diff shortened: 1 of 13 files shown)


bassosimone:
@pboothe, thanks for your review! I have been seriously swamped lately because the upcoming end of my fellowship, so I could not add the requested changes yet. `-EAGAIN`. Will retry tomorrow! :-)

bassosimone:
@pboothe I have implemented you suggestions, please take a look!

html/libndt7.js
```diff
@@ -46,13 +52,31 @@ const libndt7 = (function () {
...
+          if (settings.meta.hasOwnProperty(key)) {
+            const re = /[0-9A-Za-z._]+/
+            if (!key.match(re)) {
+              throw 'Key does not match the expected regular expression: ' + key
+            }
+            let value = settings.meta[key]
+            if (typeof value === 'number' || typeof value === 'boolean') {
+              value += ''  // force conversion to string
+            }
+            if (typeof value !== 'string' || !value.match(re)) {
```
pboothe: I think we should be okay with unicode values and that it is okay to remove the `value.match(re)` check now that we can be sure that the URLencoding will be okay in all circumstances.

pboothe: Whether or not unicode keys make sense, I leave to you. But these checks feel redundant to me.

pboothe:
Then changes look good, although perhaps things can be even 4 lines simpler in the JS?  However you decide to go, this change looks good to me.

html/libndt7.js
bassosimone: So, my understanding is that [strings in JavaScript are UTF-16](https://kev.inburke.com/kevin/node-js-string-encoding/). Reading [the polyfill code for URLSearchParams](https://github.com/WebReflection/url-search-params/blob/master/src/url-search-params.js), I additionally understand that it _should_ call `encodeURIComponent`. In turn, that [should serialise to UTF-8](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent).

Is my understanding correct?

I am concerned that JavaScript strings are UTF-16 while Go strings are UTF-8. So, I would like to be sure that removing the checks enforcing ASCII we _may_ end up with URLs that we cannot use in Go.

(This is probably just me being paranoid and not knowing enough about the Web.)

bassosimone: As a clarification, I am not thinking about using such params to build file names. But my code in another branch is serialising such params to JSON as part of saving results. Hence my concern before applying the objective improvement that you suggest. (As a note to self, if I do that, I shall also update the spec.)

pboothe:
```go
package main

import (
	"encoding/json"
	"fmt"
	"log"
	"net/url"
	"strings"
)

func main() {
	m, err := url.ParseQuery(`x=1&y=2&w=%D1%88%D0%B5%D0%BB%D0%BB%D1%8B&y=3;z`)
	if err != nil {
		log.Fatal(err)
	}
	fmt.Println(toJSON(m))
}

func toJSON(m interface{}) string {
	js, err := json.Marshal(m)
	if err != nil {
		log.Fatal(err)
	}
	return strings.Replace(string(js), ",", ", ", -1)
}
```

results in:

```
{"w":["шеллы"], "x":["1"], "y":["2", "3"], "z":[""]}
```

So it looks like UTF16 bytes are handled by Go just fine.

bassosimone: Excellent, thanks! So, I'll remove these four lines and merge :-)

PR #41, opened 2018-12-06
Refactor ndt-server into a small main and a large library

pboothe:
Creates a directory ndt to service legacy clients.

As we get to the launch of the new platform, code quality in ndt-cloud becomes more important. This is the first step of the much-larger "refactor ndt-cloud legacy into small sensible pieces" story, but it makes nice progress towards that goal.  This PR adds no functionality, it "just" moves code around.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-cloud/41)


bassosimone:
Do I understand correctly that the purpose of doing several rounds of divide et impera on the original file is to hide complexity and only export minimal interface between modules? Or is this being done for testability mainly? Or both?

pboothe:
It is being done to enable what we in google call "code velocity". 

Once this is refactored, the legacy NDT code will be 
* more easily made testable, 
* more modular with clearly-defined dependencies,
* could be made to support non-TLS clients, 
* could be made to support non-ws clients, and
* could maybe even be made to support non-JSON clients.

As it was, I did not see how to add those features easily. Now that I have cleaned this up and made explicit what depends on what, I can add tests and will feel safe that I haven't broken things when I add these features. Thus, by cleaning up first (and by adding unit tests in a future PR), I will have increased the speed at which the code can be changed.

PR #52, opened 2019-01-04
The server now speaks non-ws NDT with JSON messages.

pboothe:
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-cloud/52)

### DIFF
--- a/legacy/c2s/c2s.go
+++ b/legacy/c2s/c2s.go
@@ -106,7 +106,7 @@ func ManageTest(ws protocol.Connection, config *testresponder.Config) (float64, 
 
 	done := make(chan float64)
 	go func() {
-		// Wait for test to run. ///////////////////////////////////////////
+		// Wait for test to run.
 		// Send the server port to the client.
 		protocol.SendJSONMessage(protocol.TestPrepare, strconv.Itoa(testResponder.Port), ws)
 		c2sReady := <-testResponder.Response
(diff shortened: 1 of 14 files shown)


pboothe:
Still WIP 

When it becomes reviewable, it will have reviewers assigned. :)

legacy/c2s/c2s.go
```diff
@@ -31,19 +31,22 @@ func (tr *Responder) TestServer(w http.ResponseWriter, r *http.Request) {
...
+	go func() {
+		// After the test is supposedly over, let the socket drain a bit to not
+		// confuse poorly-written clients by closing unexpectedly when there is still
+		// buffered data. We make the judgement call that if the clients are so poorly
+		// written that they still have data buffered after 5 seconds and are confused
+		// when the c2s socket closes when buffered data is still in flight, then it
+		// is okay to break them.
+		ws.DrainUntil(time.Now().Add(5 * time.Second))
+		ws.Close()
+	}()
```
bassosimone: 👍

pboothe:
___
*[legacy/s2c/s2c.go, line 25 at r1](https://reviewable.io/reviews/m-lab/ndt-server/52#-LZXN4KT5ingqq5iwXUI:-LZeOiuY8DpQS13ff-zx:b1pfs7w) ([raw file](https://github.com/m-lab/ndt-server/blob/5971bb5a1871f69b52e6e0aa14a4ddaa5f56c875/legacy/s2c/s2c.go#L25)):*
<details><summary><i>Previously, stephen-soltesz (Stephen Soltesz) wrote…</i></summary><blockquote>

Were the types here totally wrong before? This change must impact clients that parse the value, no? A rate in json quotes is very different from not.
</blockquote></details>

The types were wrong before.  The old server code to generate the JSON is: https://github.com/m-lab/ndt/blob/master/src/test_s2c_srv.c#L625

It turns out that the `web100clt` binary expects them to be strings, and javascript clients will silently convert between strings and numbers because javascript is like that.  Computers are terrible.

PR #57, opened 2019-01-19
ndt7 spec clarifications: half duplex and minor details

bassosimone:
<strike>This is a set of [stacked diffs](https://jg.gg/2018/09/29/stacked-diffs-versus-pull-requests/). Some of them are easy fixes, some of them propose changes in the spec. What I'd like the reviewer to do is the following: (1) review the easy fixes and authorise me to merge them, if they're okay; (2) discuss about the more complex changes (i.e. using a vector of objects for measurements).</strike>

(it turns out that I didn't fully understand stacked diffs.)

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/57)

### DIFF
--- a/spec/ndt7-protocol.md
+++ b/spec/ndt7-protocol.md
@@ -102,7 +102,7 @@ server because we avoid the `TIME_WAIT` TCP state.
 If TCP BBR is available, a compliant server MAY choose to enable it
 for the client connection and terminate the download test early when
 it believes that BBR parameters become stable. Before v1.0 of this
-spec is out, we'll specify a mechanism allowing a client to opt out
+spec is out, we hope to specify a mechanism allowing a client to opt out
 of terminating the download early.
 
 Clients can detect whether BBR is enabled by checking whether the measurement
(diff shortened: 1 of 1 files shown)


spec/ndt7-protocol.md
```diff
@@ -62,12 +62,18 @@ NOT be smaller than 1 << 17 bytes.
...
 messages will contain serialised JSON structures containing measurements
-results (see below). When downloading, the server is expected to send
-measurement to the client, and when uploading, conversely, the client is
-expected to send measurements to the server. Both parties SHOULD NOT
-send measurement results when they're not expected to. Measurements SHOULD
-NOT be sent more frequently than every 250 ms, to avoid generating too
-much unnecessary processing load on the receiver.
+results (see below). Both the client and the server MAY send measurements
+messages during any subtest. Upon receiving an unexpected measurement
+message, a party SHOULD discard such message and SHOULD continue
```
evfirerob: How is `unexpected measurement` defined here? Both client and server may send measurements at any time, and the case "messages are sent too quickly" is detailed separately below. Maybe it should be `malformed message`?

bassosimone: Right, this is ambiguous. Let me address it.

bassosimone: I've pushed better text that should explain exactly what would happen. Could you read the whole diff again and let me know if there is additional ambiguity? (When we concur that the text is okay, then a bigger job is probably to have a conversation on whether we want the WebSocket to be full or half duplex during a ndt7 experiment.)

bassosimone:
Rewritten the spec changes to clarify we're half duplex!

PR #60, opened 2019-02-05
Adds UUID support to NDT7

pboothe:
Makes ndt7 save its data in a well-formed directory structure.

Also adds UUIDs to the filename and to the metadata.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/60)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,9 +10,13 @@ before_script:
 - npm install --verbose ws minimist
 
 script:
-- docker build -f TestDockerfile -t m-lab/ndt-server-test --build-arg DIRECTORY=`pwd` .
-- docker run -it -v `pwd`:`pwd` -w `pwd` m-lab/ndt-server-test
-- $GOPATH/bin/goveralls -coverprofile=ndt.cov -service=travis-ci
+- docker build -f TestDockerfile -t ndt-server-test --build-arg DIRECTORY=`pwd` .
+- docker run -t
+    -e TRAVIS_PULL_REQUEST=$TRAVIS_PULL_REQUEST
+    -e TRAVIS_JOB_ID=$TRAVIS_JOB_ID
+    -e TRAVIS=true
+    -e COVERALLS_SERVICE_NAME=travis-ci
+    ndt-server-test
 - docker build . -f Dockerfile -t m-lab/ndt-server
 
 # Need a deployment section which calls "docker push"
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ language: go
 go:
 - "1.11"
 
+dist: xenial
+
 before_script:
 - go get golang.org/x/tools/cmd/cover
 - go get github.com/mattn/goveralls
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,15 +2,17 @@ language: go
 go:
 - "1.11"
 
-dist: xenial
+services:
+- docker
 
 before_script:
-- go get golang.org/x/tools/cmd/cover
 - go get github.com/mattn/goveralls
-- go get gopkg.in/m-lab/pipe.v3
 - npm install --verbose ws minimist
 
 script:
-- go get -v -tags netgo ./cmd/ndt-cloud-client
-- go test -v -coverprofile=ndt.cov -coverpkg=./... -tags netgo ./...
+- docker build -f TestDockerfile -t m-lab/ndt-server-test --build-arg DIRECTORY=`pwd` .
+- docker run -it -v `pwd`:`pwd` -w `pwd` m-lab/ndt-server-test
 - $GOPATH/bin/goveralls -coverprofile=ndt.cov -service=travis-ci
+- docker build . -f Dockerfile -t m-lab/ndt-server
+
+# Need a deployment section which calls "docker push"
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ go:
 before_script:
 - go get golang.org/x/tools/cmd/cover
 - go get github.com/mattn/goveralls
+- go get gopkg.in/m-lab/pipe.v3
 - npm install --verbose ws minimist
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,12 +5,8 @@ go:
 services:
 - docker
 
-before_script:
-- go get github.com/mattn/goveralls
-- npm install --verbose ws minimist
-
 script:
-- docker build -f TestDockerfile -t ndt-server-test --build-arg DIRECTORY=`pwd` .
+- docker build -f TestDockerfile -t ndt-server-test .
 - docker run -t
     -e TRAVIS_PULL_REQUEST=$TRAVIS_PULL_REQUEST
     -e TRAVIS_JOB_ID=$TRAVIS_JOB_ID
--- a/build.sh
+++ b/build.sh
@@ -1,10 +1,11 @@
 #!/bin/sh
-# Script to build ndt-server with the correct `go get` flags.
+# Script to build ndt-server with the correct `go get` flags.  This script
+# should be run inside a container.
 set -ex
+
 TOPDIR=`cd $(dirname $0) && pwd -P`
 cd $TOPDIR
 VERSION=`git describe --tags`
-GOPATH=$(pwd | sed -e "s#/go/.*#/go#")
 go get -v                                                              \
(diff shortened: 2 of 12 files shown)


fdcache/fdcache.go
```diff
@@ -57,10 +57,14 @@
 package fdcache
 
 import (
+	"fmt"
+	"log"
```
bassosimone: Can you use `apex/log` rather than stdlib's `log`?

fdcache/fdcache.go
```diff
@@ -155,3 +159,17 @@ func GetAndForgetFile(conn net.Conn) *os.File {
...
+func GetUUID(conn net.Conn) (string, error) {
+	key := makekey(conn)
+	mutex.Lock()
+	defer mutex.Unlock()
+	entry, found := cache[key]
+	if !found {
+		return "", fmt.Errorf("fd not found")
+	}
+	id, err := uuid.FromFile(entry.Fp)
+	log.Println("UUID:", id)
```
bassosimone: This should be `log.Debugf("UUID: %s", id)` where `log` is `apex/log` (or, if it is just meant as a debugging aid, it should be removed before merging).

fdcache/fdcache.go
```diff
@@ -57,10 +57,14 @@
...
 
 import (
+	"fmt"
+	"log"
 	"net"
 	"os"
 	"sync"
 	"time"
+
+	"github.com/m-lab/go/uuid"
```
bassosimone: FWIW, I've also reviewed this package and found it reasonable.

fdcache/fdcache.go
```diff
@@ -155,3 +159,17 @@ func GetAndForgetFile(conn net.Conn) *os.File {
...
 }
+
+// GetUUID returns the UUID for a passed-in connection.
+func GetUUID(conn net.Conn) (string, error) {
+	key := makekey(conn)
+	mutex.Lock()
+	defer mutex.Unlock()
+	entry, found := cache[key]
+	if !found {
+		return "", fmt.Errorf("fd not found")
```
bassosimone: Could you make an `error` here using `errors.New`? (What is the difference between making an error using `errors.New` and using `fmt.Errorf`?)

ndt-server_test.go
```diff
@@ -112,30 +125,41 @@ func Test_MainIntegrationTest(t *testing.T) {
...
 
+	// Get the datadir
+	dataDir := os.Getenv("DATADIR")
+
 	tests := []struct {
 		name string
 		cmd  string
+		// ignoreData's default value (false) will NOT ignore whether data is
+		// produced. This is good, because it forces tests which ignore their output
+		// data to explicitly specify this fact.
```
bassosimone: Yeah, this is good

ndt7/server/results/file.go
```diff
@@ -16,15 +18,19 @@ type File struct {
...
-	timestamp := time.Now().UTC().Format(format)
-	name := "ndt7-" + timestamp + ".jsonl.gz"
+func newFile(datadir, what, uuid string) (*File, error) {
+	timestamp := time.Now().UTC()
+	dir := path.Join(datadir, what, timestamp.Format("2006/01/02"))
+	err := os.MkdirAll(dir, 0755)
+	if err != nil {
+		return nil, err
+	}
+	name := dir + "/" + timestamp.Format("T15:04:05.000000000Z") + "." + uuid + ".jsonl.gz"
```
bassosimone: Could you keep `ndt7-` in front?

bassosimone: Could you keep the full timestamp in the file name? (IIRC this is how I was told to do for Neubot and is _also_ much more handy if you scrape the files locally in a single directory &mdash; provided you don't have `ext2`)

ndt7/server/results/file.go
```diff
@@ -16,15 +18,19 @@ type File struct {
...
 
 // newFile opens a measurements file in the current working
 // directory on success and returns an error on failure.
-func newFile() (*File, error) {
-	format := "2006-01-02T15:04:05.000000000Z"
-	timestamp := time.Now().UTC().Format(format)
-	name := "ndt7-" + timestamp + ".jsonl.gz"
+func newFile(datadir, what, uuid string) (*File, error) {
+	timestamp := time.Now().UTC()
+	dir := path.Join(datadir, what, timestamp.Format("2006/01/02"))
```
bassosimone: Why do you separate download and upload at the dir level? (I'd have put that in the filename, perhaps)

ndt7/server/results/file.go
```diff
@@ -38,21 +44,27 @@ func newFile() (*File, error) {
...
-	initMetadata(&meta, conn.LocalAddr().String(),
-		conn.RemoteAddr().String(), request.URL.Query(), "download")
-	resultfp, err := newFile()
+	netConn := conn.UnderlyingConn()
+	id, err := fdcache.GetUUID(netConn)
+	if err != nil {
+		return nil, err
+	}
+	initMetadata(&meta, conn.LocalAddr().String(), conn.RemoteAddr().String(), id,
+		request.URL.Query(), "download")
```
bassosimone: If I understand your plan correctly, `s/"download"/what/`

pboothe:
Please take another look.

Also, avoiding the phenomenon of "actually, I no longer understand why this integration test used to pass" is a major motivation behind the development of unit tests to cover "already tested" code.  Integration tests can frequently pass "by accident" when multiple bugs happen to work together just right.

pboothe:
Travis also doesn't support socket cookies (a kernel feature from 2017). So creating files full of data is largely a non-starter because they can't be named because a UUID can't be generated.

.travis.yml
```diff
@@ -2,12 +2,17 @@ language: go
...
 script:
-- go get -v -tags netgo ./cmd/ndt-cloud-client
-- go test -v -coverprofile=ndt.cov -coverpkg=./... -tags netgo ./...
-- $GOPATH/bin/goveralls -coverprofile=ndt.cov -service=travis-ci
+- docker build -f TestDockerfile -t ndt-server-test .
+- docker run -t
+    -e TRAVIS_PULL_REQUEST=$TRAVIS_PULL_REQUEST
+    -e TRAVIS_JOB_ID=$TRAVIS_JOB_ID
+    -e TRAVIS=true
+    -e COVERALLS_SERVICE_NAME=travis-ci
```
bassosimone: IIRC also the `TRAVIS_BRANCH` variable is useful. If coverage is being computed, also `CODECOV_TOKEN` or `COVERALLS_TOKEN` (IIRC) _should_ probably be forwarded to the container. However, perhaps the `COVERALLS_SERVICE_NAME` serves the same purpose here?

build.sh
```diff
@@ -1,6 +1,8 @@
 #!/bin/sh
-# Script to build ndt-server with the correct `go get` flags.
+# Script to build ndt-server with the correct `go get` flags.  This script
+# should be run inside a container.
```
bassosimone: Why should?

test.sh
```diff
@@ -0,0 +1,13 @@
...
+set -ex
+
+# Test the NDT binary
+PATH=${PATH}:${GOPATH}/bin
+if [[ -z ${TRAVIS_PULL_REQUEST} ]]; then
+  go test -v -cover=1 -coverpkg=./... -tags netgo ./...
+else
+  go test -v -coverprofile=ndt.cov -coverpkg=./... -tags netgo ./...
+  /go/bin/goveralls -coverprofile=ndt.cov -service=travis-ci
+fi
```
bassosimone: Can you explain to me why there are two difference strategies here?

pboothe: I felt we should only send data to coveralls.io when the tests are run on Travis.

.travis.yml
pboothe: I mostly just used an existing example that worked. It looks like coveralls and travis are both happy with the current set of flags?

bassosimone: Definitely.

build.sh
pboothe: Hm. How about: "This script is designed and tested to run as part of the container build process"?

It's not forbidden to just run it, but I don't want to make any promises either.

pboothe:
Updated comments in response to your comments. :)

test.sh
bassosimone: That's reasonable.

bassosimone:
@pboothe :shipit:

PR #62, opened 2019-02-21
Set of changes: no C code, not always float64, sketch out upload

bassosimone:
This is (this time for real) a set of stacked diffs. I'd like to get a review and ack/nack for applying to master the changes that (1) get rid of C code and (2) use float64 only when it makes sense. I'd like to see some more hammering of the upload code and eventually see it merged. Whether all of this happens as part of this unique PR or multiple PRs will be created, this does not matter that much to me.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/62)


ndt7/server/upload/upload.go
```diff
@@ -0,0 +1,221 @@
...
+}
+
+// gatherAndSaveTCPInfoAndBBRInfo gathers TCP info and BBR measurements from
+// |sockfp| and stores them into the |measurement| object as well as into the
+// |resultfp| file. Returns an error on failure and nil in case of success.
+func gatherAndSaveTCPInfoAndBBRInfo(measurement *model.Measurement, sockfp *os.File, resultfp *results.File) error {
+	bbrinfo, err := bbr.GetMaxBandwidthAndMinRTT(sockfp)
+	if err == nil {
+		measurement.BBRInfo = &bbrinfo
+	}
```
bassosimone: I think it's surprising to gather BBR stats for the upload

ndt7/server/upload/upload.go
```diff
@@ -0,0 +1,221 @@
...
+	return nil
+}
+
+// measuringLoop is the loop that runs the measurements in a goroutine.
+// This function exits when
+//     (1) a fatal error occurs or
+//     (2) the maximum elapsed time for the upload test expires.
+// Since this function has access to BBR stats (if available), it's the
+// right place to stop the test early. The rest of the upload code is supposed
```
bassosimone: The part of the comment mentioning BBR should be removed.

ndt7/server/upload/upload.go
```diff
@@ -0,0 +1,221 @@
...
+// to stop receiving when this function signals that we're done by closing
+// the channel.
+func measuringLoop(ctx context.Context, request *http.Request,
+	conn *websocket.Conn, dataDir string, dst chan model.Measurement) {
```
bassosimone: My eyes bleed because of the default `go fmt` indentation when splitting lines. (Just a random comment; when we'll refactor we will probably want to have less params here, that would be the proper fix I guess.)

ndt7/server/upload/upload.go
```diff
@@ -0,0 +1,221 @@
...
+	logging.Logger.Debug("Starting measurement loop")
+	defer logging.Logger.Debug("Stopping measurement loop")
+	defer close(dst)
+
+	resultfp, err := results.OpenFor(request, conn, dataDir, "upload")
+	if err != nil {
+		return // error printed already
+	}
+
+	defer iox.ErrorLoggingCloser(resultfp).Close()
```
bassosimone: TBH I find this line of code less obvious than just `defer resultfp.Close`. This also begs the question: do we care about not being able to close the file. AFAIK, closing a file fails if there is buffered data to be committed. Should we flush the file explicitly instead and bail in case of error? Should we fail hardly if we cannot write on the disk? (I think this either means wrong configuration or disk full and in both cases we want to die.)

ndt7/server/upload/upload.go
```diff
@@ -0,0 +1,221 @@
...
+
+	sockfp, err := getConnFileAndPossiblyEnableBBR(conn)
+	if err != nil {
+		return // error printed already
+	}
+
+	defer iox.ErrorLoggingCloser(sockfp).Close()
```
bassosimone: As an additional comment to this API, I find the name not super clear. I tend to think about logging stuff as less relevant and I may inadvertently zap log related stuff, while I usually religiously keep the `defer foo.Close()` bits.

ndt7/server/upload/upload.go
```diff
@@ -0,0 +1,221 @@
...
+			if elapsed > defaultDuration {
+				logging.Logger.Debug("Upload has run for long enough.")
+				return
+			}
+
+			measurement := model.Measurement{
+				Elapsed: elapsed.Seconds(),
+			}
+
+			err = gatherAndSaveTCPInfoAndBBRInfo(&measurement, sockfp, resultfp)
```
bassosimone: I'd rename this function as here we don't need to save BBRInfo

ndt7/server/upload/upload.go
```diff
@@ -0,0 +1,221 @@
...
+// and emits Measurement events on the returned channel.
+func startMeasuring(ctx context.Context, request *http.Request, conn *websocket.Conn, dataDir string) chan model.Measurement {
+	dst := make(chan model.Measurement)
+	go measuringLoop(ctx, request, conn, dataDir, dst)
+	return dst
+}
+
+// ServeHTTP handles the upload subtest.
+func (ul Handler) ServeHTTP(writer http.ResponseWriter, request *http.Request) {
+	// TODO: factor out this stuff as it's the same for both UL/DL
```
bassosimone: I suggest to write `TODO(evfirerob)` otherwise TODOs become anti-commons

ndt7/server/upload/upload.go
```diff
@@ -0,0 +1,221 @@
...
+	headers := http.Header{}
+	headers.Add("Sec-WebSocket-Protocol", spec.SecWebSocketProtocol)
+
+	conn, err := ul.Upgrader.Upgrade(writer, request, headers)
+	if err != nil {
+		warnAndClose(writer, "Cannot UPGRADE to WebSocket")
+		return
+	}
+
+	defer conn.Close()
```
bassosimone: So, this would probably be the real place where you want to see an error from `Close`. The above `sockfp` is actually only used tangentially to access TCPInfo.

ndt7/server/upload/upload.go
```diff
@@ -0,0 +1,221 @@
...
+	// Read limit is set to the smallest allowed payload size.
+	conn.SetReadLimit(spec.MinMaxMessageSize)
+
+	// Start measuring loop
+	logging.Logger.Debug("Starting measuring goroutine")
+	ctx, cancel := context.WithCancel(request.Context())
+	measurements := startMeasuring(ctx, request, conn, ul.DataDir)
+
+	// Make sure we clean up resources
+	defer func() {
```
bassosimone: We may consider wanting to put here the `conn.Close()` and check the error here.

ndt7/server/upload/upload.go
```diff
@@ -0,0 +1,221 @@
...
+				return // the goroutine told us it's time to stop running
+			}
+		default:
+			conn.SetReadDeadline(time.Now().Add(defaultTimeout))
+			mt, message, err := conn.ReadMessage()
+			if err != nil {
+				if !websocket.IsCloseError(err, websocket.CloseNormalClosure) {
+					logging.Logger.WithError(err)
+				}
+				logging.Logger.Debug(request.RemoteAddr + ": connection closed.")
```
bassosimone: I believe this log message is redundant as we already log in the `defer func()` above

ndt7/server/upload/upload.go
```diff
@@ -0,0 +1,221 @@
...
+				break
+			}
+
+			if mt == websocket.TextMessage {
+				var mdata model.Measurement
+				err := json.Unmarshal(message, &mdata)
+				if err == nil {
+					logging.Logger.Errorf("Unable to unmarshal JSON message: %s", message)
```
bassosimone: I think we should fail the test if we receive a JSON message that we cannot parse

ndt7/server/upload/upload.go
```diff
@@ -0,0 +1,221 @@
...
+				}
+				logging.Logger.Debugf("Received Measurement - AppInfo.NumBytes: %d", mdata.AppInfo.NumBytes)
```
bassosimone: This will probably clog the log. We should remove it before merging.

ndt7/server/upload/upload.go
bassosimone: I have some additional thoughts on this `iox.ErrorLoggingCloser` API. I really don't like it (sorry). I think it makes the code more obscure and gives us a false sense of security. IMHO the code review guidlines that we'll adopt _inter nos_ should be such that we'll always ask the question whether ignoring the error of a `defer foo.Close()` is sensible or not.

When ignoring the error is sensible, we should not care much about logging. When we should not ignore the error because it indicates a specific error condition that we don't want to ignore, we should write the code such that we handle the error in some way.

Take the specific case of the disk full that I mentioned above as an example. If we cannot write results on disk, this is a serious error that we want to address, IMHO shutting down the server. Logging is not going to help in this case since we are going to be writing on a full disk anyway.

Take instead the case where we get an error when closing the connection. Do we care about writing it into the logs? I am not sure. In such case, it would probably be nicer (but way more convolute) to record in the measurement that we could not successfully close the connection. There are middleboxes around. They may mess with our TLS. So, if we want to diagnose weird cases by looking into the data, the right place where to write such information is probably into the results and not into a log file that is only local to the machine.

Basically, my argument is that you want the information to be actionable or you don't want the information. I think the logs are not the right place where to put most information. To me, logs are useful when you're developing but I do not see them much useful at scale.

(I reckon I ended up having some soapbox against this API. So, I tried to spend quite some time to explain my reasoning and I am very curious to see what others think and their counter-arguments.)

bassosimone:
Also, we should probably squash @evfirerob's upload commits before merging, so we have an history that is more bisection friendly where all commits build cleanly.

ndt7/server/upload/upload.go
pboothe: You are half right: logs are, usually, a garbage place to put information.

However, in the new world with Kubernetes our logs are on track to becoming searchable on a worldwide basis.

I like the `ErrorLoggingCloser` in principle, but I have been thinking more about it and I maybe kind of want to change its api to make the line be:

```go
	defer iox.ErrorLoggingCloser(sockfp, "Warning: ignored error").Close()
 ```

Then we can grep our logs to see whether ignored errors actually occur.  We could also (maybe) add prometheus metrics to the ErrorLoggingCloser that could make the problems available within prometheus.  (If this were a google-internal system, we would make this code generate logs at the "WARNING" level and also have the equivalent of a prometheus CounterVec counting the ignored close errors for each context. That intuitively feels like it might be overkill here, but maybe not?)

Basically, you are right that in the past logs were dumb. We are building a system where logs are not so dumb. Your (and my) old instinct that logs aren't useful in production is soon going to become obsolete. We should no longer feel like logging things to stderr/stdout is equivalent to throwing them in the trash.

Comment (bassosimone) [reply ndt7/server/upload/upload.go]: Thanks, @pboothe. I think the availability of searchable logs is a good argument that I did not consider when thinking about the problem. This clearly makes the log actionable, which is what concerned me regarding the usage.

In this vein, yeah, a specific error message can probably be helpful. Since `apex/log` emits JSON logs as configured in `ndt7`, it may perhaps make sense to emit a warning using `apex/log` also in this case.

WRT Prometheus, I'm glad to see I guessed half the right thing to do. With searchable logs, this is probably less important in general. (I think in the specific case, we probably want to know when we cannot write to disk, and maybe an alert based on that is more graceful than `panic`.)

Speaking of the API itself, I think as a programmer not super familiar with the code base, I would probably be less surprised by this alternative API:

```Go
    defer iox.CloseAndWarnOnFailure(sockfp, "ignoring sockfp.close error")
```

because it seems to me that conveys more clearly the programmer intent. I guess that, if I see a function named like this, I would not bother with searching for the definition of it, because it is self documenting. OTOH, `iox.ErrorLoggingCloser` chained with `.Close` looks a bit more obscure to me because I need to figure out what is the return value, and I would probably need to go and read the documentation when reading the code.

ndt7/server/upload/upload.go
pboothe: That would definitely be the proper fix.

ndt7/server/upload/upload.go
```diff
@@ -0,0 +1,206 @@
+package upload
```
bassosimone: I believe a docstring is missing here. Can you check that using `golint`?

ndt7/server/upload/upload.go
bassosimone: I am going to close this conversation as redundant.

ndt7/server/upload/upload.go
```diff
@@ -0,0 +1,206 @@
...
+// and emits Measurement events on the returned channel.
+func startMeasuring(ctx context.Context, request *http.Request, conn *websocket.Conn, dataDir string) chan model.Measurement {
+	dst := make(chan model.Measurement)
+	go measuringLoop(ctx, request, conn, dataDir, dst)
+	return dst
+}
+
+// ServeHTTP handles the upload subtest.
+func (ul Handler) ServeHTTP(writer http.ResponseWriter, request *http.Request) {
+	// TODO(evfirerob): factor out this stuff as it's the same for both UL/DL
```
bassosimone: (We don't know for sure until the code in here has taken its final shape. Let's remember this.)

ndt7/server/upload/upload.go
bassosimone: Second thinking, this is not a good pattern. The best pattern is to `defer conn.Close()` as soon as you know it's a valid instance and not `nil`. Saying this because putting the closing of the connection inside of the func is conducive to possible leaks if you don't notice and introduce some new code between the point where the conn is valid and the point in which the func is deferred. This vouches in favour of the approach that I criticised earlier of having some way of knowing if you're ignoring a close. (I still think that, when you know you don't want that to be an error, you probably want an explicitly deferred lambda with proper error checking, however the `iox` interface is a sled between doing nothing and going all in with a full lambda.)

Do you, @evfirerob, or @pboothe have any counter arguments or objections? Otherwise, we can probably resolve this branch of the conversation and move forward.

ndt7/server/upload/upload.go
bassosimone: I think, @evfirerob, you should comment on this

evfirerob: I fixed the logic _and_ added a break in case the JSON message is unparseable.

pboothe:
___
*[ndt7/server/upload/upload.go, line 179 at r4](https://reviewable.io/reviews/m-lab/ndt-server/62#-LZF7vKzSljLAwe1GaEX-r4-179:-LZoX8a3-KlTFNU6vDi3:b2wv54) ([raw file](https://github.com/m-lab/ndt-server/blob/49359248bd6bfc6b6f5975c91bbbb95b2b422afa/ndt7/server/upload/upload.go#L179)):*
<details><summary><i>Previously, bassosimone (Simone Basso) wrote…</i></summary><blockquote>

Second thinking, this is not a good pattern. The best pattern is to `defer conn.Close()` as soon as you know it's a valid instance and not `nil`. Saying this because putting the closing of the connection inside of the func is conducive to possible leaks if you don't notice and introduce some new code between the point where the conn is valid and the point in which the func is deferred. This vouches in favour of the approach that I criticised earlier of having some way of knowing if you're ignoring a close. (I still think that, when you know you don't want that to be an error, you probably want an explicitly deferred lambda with proper error checking, however the `iox` interface is a sled between doing nothing and going all in with a full lambda.)

Do you, @evfirerob, or @pboothe have any counter arguments or objections? Otherwise, we can probably resolve this branch of the conversation and move forward.
</blockquote></details>

TL;DR: no objections.

You should `defer` any cleanup code immediately upon being certain that the cleanup code won't crash. Not only does it prevent new code from "sneaking in" the middle, but it means that every open has, by inspection, a corresponding close, because you can see the close on the next line.

Put another way: many small `defer` statements are preferable to one big one.

ndt7/server/upload/upload.go
pboothe: I think that third party API users are the ultimate authority on what the most readable solution is. So if you think that `CloseAndWarnOnFailure` is clearer, then so do I, as you are the closest thing we currently have to a user of the library.

Here is one last idea to stir the pot:  What if we called the library `warnonfailure` instead of `iox`? Then the code would be:

```go
defer warnonfailure.Close(sockfp, "ignoring sockfp.Close error")
```

It also suggests that other methods could be added to the library, which would more closely mirror Roberto's original intent.

bassosimone: Yes, I think changing the name as you propose is great and makes the code super clear and explicit! Thanks!

bassosimone: Resolving since @evfirerob implemented this suggestion.

bassosimone:
Supeseded by #99

Issue #75, opened 2019-03-08
ndt7: spec should clarify the closing protocol


bassosimone:
<strike>The sane closing protocol is that the source (i.e. the server during the download) sends a CloseMessage when done, then the sink (i.e. the client during the download) finshes uploading its counter-flow measurements, if any, and then also sends a CloseMessage. This shall unblock the server reader and then terminate the test. Note that this also poses some constraints on how the code is written.

(Learned this the hard way when writing the reference Go client for ndt7.)</strike>

Well, actually the correct thing to do is to write code that is fault tolerant with respect to the closing protocol, and does not declare failure once the test is started. This is especially useful because either end should be able to RST the connection to stop the peer, as a last resort solution, and the test should still be declared successful in this case.

So, there is still need to document the closing protocol, but differently than originally planned.

[Edited on 29 August 2019]

bassosimone:
This is basically another angle of #72, so I'm going to close this as duplicate.

PR #82, opened 2019-03-11
Hammer the download until it looks good

bassosimone:
The description below is now historical. Since the [original review by @pboothe](https://github.com/m-lab/ndt-server/pull/82#pullrequestreview-215711873), I continued working on the code until the semantics issues concerning the usage of pipelines were solved.

- - -

This PR refactors the download code to be a pipeline and to implement the closure procotol described in #75, which is working with prototype half-duplex and full-duplex clients.

The structure of the pipeline (to be better documented in a PR) is the following:

```
conn -----.---------------.-------------.---------------.
          |               |             |               |
          V               V             V               V
       +----------+    +--------+    +----------+    +--------+
ctx -> | measurer | => | sender | => | receiver | => | closer | -> error
       +----------+    +--------+    +----------+    +--------+
               Measurement      error           error
```

where:

- *ctx* expires when either the whole request context expires or the maximum download runtime (ten seconds) expires.

- *measurer* runs until *ctx* expires, or until an error occurs, and periodically performs BBR and TCPInfo measurements, which are sent downstream towards *sender*. Of course, the downstream channel is closed when the *measurer is done*.

- *sender* loops sending data to client and intermixing such data with measurements read from upstream in nonblocking fashion. It writes on the output channel when an error occurred. This stage terminates when there is an error or the previous stage terminates (i.e. the channel is closed). It drains the input channel and closes the outut channel when done. Also, when done, it initiates the shutdown protocol (see #75) by sending a CLOSE websocket message to the peer.

- *receiver* loops until either an error occurs upstream or it's not possible to read further messages from the client, because of a network error or a normal WebSocket closure message (which should be the clean way to indicate that no more counter-flow messages follow &mdash; see #75). The upstream channel is sampled in nonblocking fashion because we want to continue reading also after the upstream stage is gone, to read all the counter-flow server-side measurements; in fact, this stage should be terminated by the peer sending us a CLOSE message to respond to our previous CLOSE message. The input channel is drained, the output channel is closed, when leaving.

- *closer* terminates after *receiver* terminates or the maximum test duration (i.e. fifteen seconds) expires. In any case, the connection is closed. In case of timeout, possibly caused by an half-duplex client, closing the connection unblocks the *receiver* stage. The input channel is always drained.

As of 2019-03-12T00:04:00 CET, this PR depends on #80 to be merged. So I am initially flagging it as a draft PR and will not change its state until #80 is merged. Then I'll make this one mergeable.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/82)


bassosimone:
Oh, replying to myself: I actually think it's possible to have a pipeless implementation with the same property, @pboothe. I think you're right that I can simplify this further. Let me try again.

bassosimone:
@pboothe I wrote a PoC of how the implementation of the download would look like when using two pipelines (one for sending and one for receiving) rather than one. It's here:

https://github.com/m-lab/ndt-server/tree/attempt

I did this so that we can compare the two implementations and discuss. This PoC has, AFAICT, the same properties of the single pipeline. It's nicer that pipelines are actually pipelines that move around data (which is probably less surprising). I am not sure which one I like more.

bassosimone:
> @pboothe I wrote a PoC of how the implementation of the download would look like when using two pipelines (one for sending and one for receiving) rather than one. It's here:
>
> https://github.com/m-lab/ndt-server/tree/attempt
>
> I did this so that we can compare the two implementations and discuss. This PoC has, AFAICT, the same properties of the single pipeline. It's nicer that pipelines are actually pipelines that move around data (which is probably less surprising). I am not sure which one I like more.

I had a chat with @pboothe in which he suggested to keep the pipeline for now along with a comment saying that we may want to improve upon that in the future. He likes the pipeline attempt more than the other PoC that I sketched out this morning.

bassosimone:
@pboothe I dismissed the previous review and submitted the code I showed you a few days ago for a new review. Please, take a look to this new changes!

bassosimone:
The tests are currently failing because the embedded client fails if it does not receive the message it exactly expects to end the test. As discussed with @pboothe, we should always remember about [1453](https://en.wikipedia.org/wiki/Byzantine_fault) when designing systems. A real world client will be more flexible, so it's fine to have a more flexible testing client. (FWIW, when running the same test locally this failure does not appear, which is great because Travis is teaching us^H^H me that I should design robust systems and additional layers of abstractions will complicate things).

So, I'll fix the client rather than the server here, and unblock the PR.

PR #85, opened 2019-03-19
Cancel -> Close

pboothe:
`Cancel` is an ephemeral member of the `TestResponder` struct.  `Close` is a permanent method that may or may not call `Cancel`.  Only `Close` should call `Cancel`.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/85)


pboothe:
image

I don't know why I kept these two things right next to each other, but in retrospect it was a bad choice.

PR #99, opened 2019-04-10
Initial implementation of ndt7 upload

bassosimone:
I think it's better to start merging an implementation of the upload, albeit imperfect, and then experiment with it while this implementation is in tree and deployed on some test servers.

Supersedes #62 

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/99)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@
 # See <https://docs.travis-ci.com/user/languages/minimal-and-generic/>
 # See <https://travis-ci.org/m-lab/ndt-server/builds/496487913>
 language: minimal
+dist: xenial
 
 services:
 - docker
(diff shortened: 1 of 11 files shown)


bassosimone:
The efb1711 ("Is my intuition correct?") commit proves that #100 and #101 are a blocker to move forward this pull request, because we clearly have issues with Travis.

pboothe:
But still no unit tests for upload.go?

pboothe:
Future work plan:
1. make it work
2. Refactor it to make it testable and add tests
3. Call the testable design "good"

This accomplishes goal 1, and Simone has agreed that he will follow up with goals 2 and 3 with 100% coverage as the design goal once all the yaks are shaved.

LGTM

Issue #100, opened 2019-04-11
ndt-server build broken on macOS


bassosimone:
A reasonable reaction to this issue may be:

> we don't support macOS

However, I contend that this is not the right thing to do. My (notoriously deprecated) intuition says that Travis is actually broken for ndt7 and this has to do with `$something`.

My intuition also says that it's seeing what happens on macOS helps to understand what happens on Travis. So, the first step is to make the code build on macOS (a PR for that is coming). The second step is understanding this error:

> {"fields":{"error":"Error in Getsockopt. Errno=42"},"level":"warn","timestamp":"2019-04-11T17:44:03.670879+02:00","message":"fdcache.GetUUID failed"}

This is the error that prevents a download from starting on macOS. According to my intuition this is also happening on Travis and fixing this would make Travis work again.

(Yeah, this issue is very mystical, I reckon that.)

bassosimone:
Ah-ha! Understanding the error is surprisingly easy. The following code in [m-lab/uuid.go](https://github.com/m-lab/uuid/blob/6bb4635a3de732b0b0c448c7ac654f909c785c30/uuid.go#L63) is non portable to systems that are not Linux:

```go
// getCookie returns the cookie (the UUID) associated with a socket. For a given
// boot of a given hostname, this UUID is guaranteed to be unique (until the
// host receives more than 2^64 connections without rebooting).
func getCookie(file *os.File) (uint64, error) {
	var cookie uint64
	cookieLen := uint32(unsafe.Sizeof(cookie))
	// GetsockoptInt does not work for 64 bit integers, which is what the UUID is.
	// So we crib from the GetsockoptInt implementation and ndt-server/tcpinfox,
	// and call the syscall manually.
	_, _, errno := syscall.Syscall6(
		uintptr(syscall.SYS_GETSOCKOPT),
		uintptr(int(file.Fd())),
		uintptr(syscall.SOL_SOCKET),
		uintptr(syscallSoCookie),
		uintptr(unsafe.Pointer(&cookie)),
		uintptr(unsafe.Pointer(&cookieLen)),
		uintptr(0))

	if errno != 0 {
		return 0, fmt.Errorf("Error in Getsockopt. Errno=%d", errno)
	}
	return cookie, nil
}
```

My initial reaction is to just avoid this codepath on non Linux systems to fix the issue for macOS but of course that will not help to verify or refute my hypothesis that also travis fails.

bassosimone:
[...]

PR #101, opened 2019-04-11
Make sure macOS works with reduced functionality

bassosimone:
We don't support macOS officially. However it's useful to build on
a system that is different from the containers where we deploy.

As mentioned in #100, I am convinced that Travis is different enough
from the containers where we deploy to look like macOS.

See also #100.

After opening the PR, I've also made sure that a test works (of
course without BBR etc etc etc)

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/101)


bassosimone:
So, this contains bits that @pboothe suggests to put in m-lab/uuid rather than here. The review can resume after I have moved this bit of code to the other repository.

PR #109, opened 2019-05-02
ndt7-protocol.md: clients mlab-ns, scheduling, and back-off requirements

### LINKED ISSUE #34: ndt7/spec: mention mlab-ns best practices
If J. Random Hacker is writing its own ndt7 client based on the spec, and considering that they are likely to want to use _at least_ also the ndt7 servers at M-Lab, it becomes very important to mention in the spec how we recommend them to use mlab-ns.

Discussed with @pboothe a while ago. Written here so we keep track.

### LINKED ISSUE #35: spec/ndt7: mention periodic testing best practices
This is a topic that I discussed in the past with @mattmathis and others in the team. Basically, we should make sure that, when J. Random Hacker is writing their own ndt7 client, they also use the correct methodology to run periodic tests. This is helpful to them and to us.

bassosimone:
This diff modifies the ndt7 specification to introduce client
requirements related to mlab-ns.

I have added more information, at the cost of possibly duplicate
some information and/or being redundant with mlab-ns docs, because
IMO this helps us be sure that people who are writing clients for
ndt7 understand how to interact with the platform.

Other solutions to the same problem may be possible and I don't
have a strong opinion that this is the best way.

Closes #34.

Closes #35.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/109)


spec/ndt7-protocol.md
```diff
@@ -192,7 +192,110 @@ implementations will likely round bigger (or smaller) numbers to the nearest
...
+is identifiable by mlab-ns returning `200` along with an empty set of servers;
+
+7. MUST use the following retry policy (if applicable): extract a random
+number of seconds from an exponential distribution with average equal
+to `3,600` seconds, and wait that number of seconds before trying again;
+
+8. MUST (of course) interpret `200` as success and any other status
+code as failure, and act accordingly.
+
+For the purpose of the beta testing phase of `ndt7`, and whenever they wish
```
pboothe: Remove "and whenever they wish to use non-production infrastructure". We don't want people using the staging MLab-NS instance.

I will see how hard it is to use the newly-created reverse proxy functionality to  forward `ndt7` requests to staging, which would eliminate the need to ever tell anyone about `https://locate-dot-mlab-staging.appspot.com`

spec/ndt7-protocol.md
```diff
@@ -192,7 +192,110 @@ implementations will likely round bigger (or smaller) numbers to the nearest
...
+ndt7-enabled servers for the purpose of running tests. Of course, this
+section only applies to clients using M-Lab infrastructure. Clients
+using other server infrastructure MUST refer to the documentation
+provided by such infrastructure instead.
+
+Clients:
+
+1. SHOULD use [mlab-ns](https://github.com/m-lab/mlab-ns);
+
+2. MUST query for the `ndt_ssl` mlab-ns tool (see below for a
```
pboothe: Should be `/ndt7`

spec/ndt7-protocol.md
```diff
@@ -192,7 +192,110 @@ implementations will likely round bigger (or smaller) numbers to the nearest
...
+until ndt7 is deployed on the production infrastructure):
+
+```
+https://mlab-ns.appspot.com
+```
+
+Thus, the full URL to contact mlab-ns MUST be:
+
+```
+${baseURL}/ndt_ssl?policy=geo_options
```
pboothe: Why these url params? They could also just ask for a single server using the bare URL.

spec/ndt7-protocol.md
```diff
@@ -192,7 +192,110 @@ implementations will likely round bigger (or smaller) numbers to the nearest
...
+return a set of nearby servers;
+
+5. if the connection to the first returned server is down, they SHOULD
+try with subsequent servers, and fail the test if all the returned
+servers appear to be down;
+
+6. MUST handle gracefully the case where M-Lab is out of capacity, which
+is identifiable by mlab-ns returning `200` along with an empty set of servers;
+
+7. MUST use the following retry policy (if applicable): extract a random
```
pboothe: retry policy is for errors. For errors we want exponential backoff. For repeated tests across time, we want a poisson process.  Which of these scenarios are you trying to give guidance for here?

spec/ndt7-protocol.md
bassosimone: Because having more than one server is more robust: a server may be down and we still want to run a test with a backup server. Isn't this the reason why `geo_policy` was added and is used by specific clients to have more robustness?

pboothe: Yes, but that was back when our systems were less reliable and mlab-ns had a 45 minute update latency.  Now it has a much tighter (under 2 minutes) freshness bound, and is serving a much more accurate view of the world. There's a lot less need for the client to need to get a list of servers unless they have very high reliability needs.

spec/ndt7-protocol.md
pboothe: 1. Yes.

2. Yes. If the test fails, I would recommend they just move on. Unless they really need a result soon, in which case exponential backoff +jitter is the right thing.  Exponential backoff is: try, wait 1 minute and retry, then wait 2 minutes and retry, then wait 4 minutes to retry, then wait 8 minutes and retry....

spec/ndt7-protocol.md
bassosimone: Thanks! I've removed that bit.

spec/ndt7-protocol.md
bassosimone: Done

spec/ndt7-protocol.md
bassosimone: Awesome: I didn't know about this and really like to avoid to loop.

spec/ndt7-protocol.md
bassosimone: Ok, tried to capture that in the text!

spec/ndt7-protocol.md
```diff
@@ -192,7 +192,158 @@ implementations will likely round bigger (or smaller) numbers to the nearest
...
+```
+< HTTP/2 200
+< cache-control: no-cache
+< access-control-allow-origin: *
+< content-type: application/json
+< date: Thu, 02 May 2019 13:46:32 GMT
+< server: Google Frontend
+< content-length: 2
+<
+{}
```
bassosimone: @stephen-soltesz is this correct? (do you think it would be possible to return `204` in this error case? I have a slight preference for using the error codes, if that's OK with all clients.)

bassosimone: This has now been done!

spec/ndt7-protocol.md
```diff
@@ -192,7 +192,158 @@ implementations will likely round bigger (or smaller) numbers to the nearest
...
+5. MUST handle gracefully the case where M-Lab is out of capacity, which
+is identifiable by locate.measurementlab.net returning `200` along with an
+empty response.
+
+The following example shows a request to locate.measurementlab.net originating
+from a well-behaved ndt7 client:
+
+```
+* Connected to locate.measurementlab.net (216.58.205.84) port 443 (#0)
+> GET /ndt7 HTTP/2
```
bassosimone: Note to self: use `HTTP/1.1`: cURL actually represents HTTP/2 in a way that looks like HTTP/1.1, but HTTP/2 is different and has pseudo-headers. So I don't want to misrepresent HTTP/2 here.

bassosimone:
@stephen-soltesz @pboothe I've updated the PR to match changed since it was opened. With the `mlab-ns` change landing soon 🤞, I think it makes sense to have the spec in the master branch updated such that it references the correct URL for mlab-ns.

bassosimone:
FTR, `https://locate.measurementlab.net/ndt7` is now WAI. So merging this PR would not mean that our specification contains fake news or non working endpoints! 🎉

bassosimone:
@pboothe can you please take another look?

Issue #107, opened 2019-04-23
NDT legacy server slowly dies when serving production traffic


pboothe:
Reported by nkinkade:
`ERROR: manageS2cTest context deadline exceeded` seems to appear a lot.

This gives new urgency to the code cleanups planned for ndt-server.
￼

pboothe:
We need to improve the server logging if we are to make any headway on this.

PR #115, opened 2019-05-14
Fix goroutine leak

pboothe:
Call the correct code.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/115)


pboothe:
Fixes #107

PR #127, opened 2019-05-31
spec: set max message size at 1<<20

### LINKED ISSUE #113: spec/ndt7: raise min message size to 1 << 24
This is probably more future proof than 1<<17

bassosimone:
Closes #113

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/127)


pboothe:
I think 1<<20 bytes (1 Megabyte messages) is going to be important for Javascript browser clients on very very fast interfaces.  For 10G interfaces, it's still 1000+ messages per second.  If anything, I think a max size of 4 or 16 Megabytes per message is a better idea.  Clients still shouldn't make their messages that big unless they know what they are doing, but the protocol should not forbid messages that are that big.

bassosimone:
See #130 for some benchmarks aimed at understanding whether larger messages help.

PR #140, opened 2019-06-27
docker: reduce privileges when running for testing

### LINKED ISSUE #141: README: run unprivileged container
I do not see compelling reasons for running the `ndt-server` container in privileged mode and we actually can run it as the current user, as far as local deployments are concerned.

I'd like this to happen quite soon, because people testing their own deployments will probably follow the README and I'd like to recommend unprivileged containers.

bassosimone:
When running locally for testing, we can reduce the privileges of the container by removing all the capabilities we don't need, by using a really minimal container, by making sure that what needs to be mounted in read-only mode is not mounted in read-write mode.

Closes #141 

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/140)


bassosimone:
@pboothe wrote:

> Maybe gcr.io/distroless/static instead?

Thanks for the suggestion, will check it out!

bassosimone:
I've reverted back to using Alpine for running, as suggested by @stephen-soltesz on Slack:
> some months ago we adopted alpine instead of scratch to enable interactive introspection and debugging of the container-visible environment. While I think we would be open to moving to scratch in the future right now when we're still debugging the ndt server on the regular, I would say it's too soon.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.12-alpine as ndt-server-build
+FROM golang:alpine as ndt-server-build
```
bassosimone: This is to track the latest version of Go when 1.13 will be out

bassosimone:
Closes #141 

Follow-up issue #142

bassosimone:
@pboothe please, take another look

Issue #70, opened 2019-03-04
ndt7/upload: the server should provide feedback to the client


bassosimone:
During the upload, the client may or may not have access to BBR stats. That is actually the common case when you are running in the browser. Such use case may benefit from the server sending back to the client every, say, 0.25 seconds, stats about the upload speed. This will provide clients with a more correct view of the real upload speed, since the one measured on client side is too optimistic because of buffering.

PR #148, opened 2019-07-09
Rewrite upload using download code

bassosimone:
This diff is yak shaving functional to implement #70. Another improvement in sight that this PR will facilitate is writing unit tests, since the code has been unified and there's less to test.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/148)


README.md
```diff
@@ -24,7 +24,7 @@ docker build . -t ndt-server
 prepare the runtime environment
 
 ```bash
-install -d certs data
+install -d certs datadir
```
bassosimone: I changed the name because `data` now contains Go code. It may be bad to suggest storing data with a default command line in a place where there is also Go code, because one may be tempted to `rm -rf` it.

ndt7/closer/closer.go
```diff
@@ -0,0 +1,23 @@
+// Package closer implements the WebSocket closer.
```
bassosimone: Factored out code to allow sharing it.

ndt7/download/receiver/receiver.go
```diff
@@ -1,63 +0,0 @@
-// Package receiver implements the counter-flow messages receiver.
```
bassosimone: Renamed (unclear to me why it's not shown as a renamed and slightly changed file)

Issue #139, opened 2019-06-26
NDT legacy does not save the port number.


pboothe:
NDT legacy does not save the port number for either the client or the server. It should save both.

PR #149, opened 2019-07-16
Unify ndt5 and ndt7 result schema

stephen-soltesz:
This change modifies the data.NDTResult struct used to serialize the test summary data for both the ndt5 and ndt7 protocols. Previously, ndt7 logs were JSONL -- now the results are a single data.NDTResult serialized as JSON.

This change introduces two new fields to the NDTResult struct for ndt7 `Upload` and `Download` tests. To change how data is written (from line-based to whole-object based) the `result.File` interface is updated to allow incrementally "saving" test `model.ArchivalData` such as client metadata, client & server measurements, and other fields.

This change also consolidates the `ndt5data.Metadata` and client-provided `Meta` fields into a single `control.ArchivalData` structure corresponding to the ndt5 "control channel" connection. Now every top-level object in the NDTResult struct represents a *connection* to the NDT server and all metadata collected by that connection.

As well, client and server ports are recorded in the result data for both protocols.

As well, the struct field names are uniform throughout, preferring "UUID" globally and "ClientMetadata" for client-provided metadata.

Part of https://github.com/m-lab/ndt-server/issues/143

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/149)


data/result.go
```diff
@@ -18,19 +20,25 @@ import (
...
+	Version string
 
 	// All data members should all be self-describing. In the event of confusion,
 	// rename them to add clarity rather than adding a comment.
-	NDT5Metadata *ndt5data.Metadata `json:",omitempty"`
-
 	ServerIP   string
 	ServerPort int
 	ClientIP   string
 	ClientPort int
```
bassosimone: By reading the code as-is, I'd assume these four describe the control channel for `ndt5` and the control+test channel for `ndt7`. Is my understanding correct?

ndt7/download/measurer/measurer.go
```diff
@@ -62,6 +62,8 @@ func loop(
 	}
 	defer sockfp.Close()
 	t0 := time.Now()
+	resultsfp.StartTest()
+	defer resultsfp.EndTest()
```
bassosimone: This is nice!

ndt7/download/measurer/measurer.go
```diff
@@ -77,10 +79,11 @@ func loop(
 			}
 			measure(&measurement, sockfp)
 			if sentConnectionInfo == false {
+				// TODO: eliminate ConnectionInfo from model.Measurement and send to client differently.
```
bassosimone: Please, rather than a comment, create an issue for this task. I believe it should be possible to do that without breaking clients, but we should do that and see. Given that roll-out is happening soon, I want to make sure we do that (or not do that) before, so that we don't break stuff after. An issue seems better to ensure that.

ndt7/handler/handler.go
```diff
@@ -65,13 +70,41 @@ func (h Handler) downloadOrUpload(writer http.ResponseWriter, request *http.Requ
 	// and close(2), we'll end up keeping sockets that caused an error in the
 	// code above (e.g. because the handshake was not okay) alive for the time
 	// in which the corresponding *os.File is kept in cache.
-	defer warnonerror.Close(conn, "download: ignoring conn.Close result")
+	defer warnonerror.Close(conn, "downloadOrUpload: ignoring conn.Close result")
```
bassosimone: Thanks!

ndt7/handler/handler.go
```diff
@@ -65,13 +70,41 @@ func (h Handler) downloadOrUpload(writer http.ResponseWriter, request *http.Requ
...
 	logging.Logger.Debug("downloadOrUpload: opening results file")
 	resultfp, err := results.OpenFor(request, conn, h.DataDir, kind)
 	if err != nil {
 		return // error already printed
 	}
-	defer warnonerror.Close(resultfp, "download: ignoring resultfp.Close result")
+
+	// Collect test metadata.
+	clientAddr := conn.RemoteAddr().(*net.TCPAddr)
+	serverAddr := conn.LocalAddr().(*net.TCPAddr)
```
bassosimone: I am slightly confused here. Will the `*net.TCPAddr` cast _also_ work for `tls.Conn` connections?

ndt7/model/archivaldata.go
```diff
@@ -0,0 +1,15 @@
+package model
+
+import (
+	"time"
+)
```
bassosimone: This can be `import "time"`

ndt7/model/archivaldata.go
```diff
@@ -0,0 +1,15 @@
...
+
+// ArchivalData saves all instantaneous measurements over the lifetime of a test.
+type ArchivalData struct {
+	UUID               string
+	StartTime          time.Time
+	EndTime            time.Time
+	ServerMeasurements []Measurement
+	ClientMeasurements []Measurement
+	ClientMetadata     map[string]string `json:",omitempty" bigquery:"-"`
```
bassosimone: This may contain useful information like the client name and version. If I understand correctly the `-` here you're saying you don't want to export this information to BigQuery. Can you explain/document why?

ndt7/results/file.go
```diff
@@ -105,27 +107,39 @@ type savedMeasurement struct {
 	// Measurement is the actual measurement to be saved.
 	Measurement model.Measurement `json:"measurement"`
 }
+*/
```
bassosimone: If this code is not needed anymore, can we zap it rather than commenting it out? If there is a reason to keep it commented out, instead, I'd suggest to explain that with a `//` comment before the `/*`.

ndt7/results/file.go
```diff
@@ -105,27 +107,39 @@ type savedMeasurement struct {
...
+
+// SaveServerMeasurement saves the |measurement| for archival data.
+func (fp *File) SaveServerMeasurement(measurement model.Measurement) {
+	fp.Data.ServerMeasurements = append(fp.Data.ServerMeasurements, measurement)
+}
+
+// SaveMetadata writes |metadata| on the measurements file.
+func (fp *File) SaveMetadata(metadata metadata) {
+	fp.Data.ClientMetadata = metadata
+}
```
bassosimone: (Maybe this is the non-native bias but) it confused me to see the `Save` verb used when data is not saved on disk but just added to a data structure. In fact, above, I find myself wondering what `SaveMetadata` did.

I'd suggest to either explain me why `save` is the right verb in this context or to use another verb. (I'd instinctively use `set` for setting a single variable and `append` for appending to a slice.)

ndt7/results/file.go
```diff
@@ -25,6 +25,9 @@ type File struct {
 
 	// UUID is the UUID of this subtest
 	UUID string
+
+	// Result contains metadata about the complete measurement.
+	Data *model.ArchivalData
```
bassosimone: The docstring should document `Data` not `Result`.

ndt7/handler/handler.go
```diff
@@ -65,13 +70,41 @@ func (h Handler) downloadOrUpload(writer http.ResponseWriter, request *http.Requ
...
+		Version:        version.Version,
+		ClientIP:       clientAddr.IP.String(),
+		ClientPort:     clientAddr.Port,
+		ServerIP:       serverAddr.IP.String(),
+		ServerPort:     serverAddr.Port,
+		StartTime:      time.Now(),
+	}
+	// Guarantee that we record an end time, even if tester panics.
+	defer func() {
+		result.EndTime = time.Now()
```
bassosimone: I am confused here because it seems you are setting `EndTime` already by `defer resultfp.EndTest` when you are in `download/measurer/measurer.go`. Also, I'd be happier if we could add more abstraction and not expose `Data` but just make it an implementation detail of `resultfp` accessible via APIs.

bassosimone: Like, ideally, here I'd invoke a constructor of initialiser of `resultfp` and then perhaps `defer resultfp.WriteBack` or something similar, rather than having this long deferred function.

ndt7/results/metadata.go
```diff
@@ -14,23 +12,13 @@ type metadata map[string]string
...
+// initMetadata initializes |*meta| from |values| provided from the original
+// request query string.
+func initMetadata(m *metadata, values url.Values) {
 	for k, v := range values {
 		if matches := serverKeyRe.MatchString(k); matches {
 			continue // We MUST skip variables reserved to the server
 		}
-		if len(v) != 1 {
-			continue // We SHOULD ignore duplicate keys
-		}
```
bassosimone: Why are you removing this code? At least you should check for `len(v) > 0` because in principle `v[0]` can panic and this string here is provided by the client.

ndt7/results/metadata.go
```diff
@@ -14,23 +12,13 @@ type metadata map[string]string
...
 		(*m)[k] = v[0]
 	}
-	(*m)["server_name"] = "ndt-server"
```
bassosimone: In a distant future we may want to have another server implementation. Does it make sense to modify the top-level data structure to also include the server name?

ndt7/upload/upload.go
```diff
@@ -50,19 +50,12 @@ func getConnFile(conn *websocket.Conn) (*os.File, error) {
 }
 
 // gatherAndSaveTCPInfo gathers TCP info measurements from |sockfp| and stores
-// them into the |measurement| object as well as into the |resultfp| file.
-// It returns an error on failure and nil in case of success.
-func gatherAndSaveTCPInfo(measurement *model.Measurement, sockfp *os.File, resultfp *results.File) error {
+// them into the |measurement| object. On error, the measurement.TCPInfo will remain unchanged.
+func gatherAndSaveTCPInfo(measurement *model.Measurement, sockfp *os.File) error {
```
bassosimone: This is now `gatherTCPInfo` because it does not write on disk, right?

ndt7/upload/upload.go
```diff
@@ -153,16 +149,17 @@ func Do(ctx context.Context, conn *websocket.Conn, resultfp *results.File) {
...
-				err := json.Unmarshal(message, &mdata)
+				var m model.Measurement
+				err := json.Unmarshal(message, &m)
 				if err != nil {
 					logging.Logger.Errorf(
 						"upload: unable to unmarshal JSON message: %s", message,
 					)
 					return
 				}
+				// Save this message from the client.
```
bassosimone: The comment is probably redundant?

spec/data-format.md
```diff
@@ -1,84 +1,92 @@
...
+## Result JSON
 
-The first line in the JSONL file contains metadata. It consists of
-an object mapping keys to string values. The keys contained in this
-JSON are the ones provided by the client in the query string plus
-the ones reserved to the server, as specified in the "Query string
-parameters" section of [ndt7-protocol.md](ndt7-protocol.md).
+The result JSON value is complete record of the test. It consists of an
+object with fields for client and server IP and port, start and end time, and
+for NDT7 either an upload or download summary data.
```
bassosimone: Can you use `ndt7` here for consistency with other docs?

spec/data-format.md
```diff
@@ -1,84 +1,92 @@
...
+    "ServerIP": "::1",
+    "ServerPort": 443,
+    "StartTime": "2019-07-16T15:26:05.987748459-04:00",
+    "EndTime": "2019-07-16T15:26:16.008714743-04:00",
+    "Upload": {
+        "StartTime": "2019-07-16T15:26:05.987853779-04:00",
+        "EndTime": "2019-07-16T15:26:16.008677965-04:00",
+        "UUID": "soltesz99.nyc.corp.google.com_1563200740_unsafe_00000000000157C6",
+        "ClientMeasurements": [
+          ...
```
bassosimone: Omitting the `...` would prevent GitHub from rendering a JSON parse error.

spec/ndt7-protocol.md
```diff
@@ -218,7 +218,7 @@ Where:
 Note that JSON and JavaScript actually define integers as `int53` but existing
 implementations will likely round bigger (or smaller) numbers to the nearest
 `float64` value. A pedantic implementation MAY want to be overly defensive and
-make sure that it does not mit values that a `int53` cannot represent. The
+make sure that it does not emit values that a `int53` cannot represent. The
```
bassosimone: Thanks!

ndt7/results/metadata.go
stephen-soltesz: Can you provide an example where v is empty? See for example: https://play.golang.org/p/NXba6pTmCNa

ndt7/results/metadata.go
stephen-soltesz: What would a top-level field for "ServerName" provide that the git commit and version fields do not?

stephen-soltesz:
Part of https://github.com/m-lab/ndt-server/issues/139

PR #155, opened 2019-07-20
ndt7/download: start adapting message size

bassosimone:
Depends on #148 

This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/155)


bassosimone:
This was a good first attempt, but I found out that there were issues with using BBRInfo for this purpose, because we were using the sampling of a sampling.

Also, we discussed with @mattmathis and @pboothe that we wanted, if possible, the core of ndt7 not to depend on BBRInfo or TCPInfo, if possible.

I'm going to close this PR and open a future one with a better strategy that we're testing.

Issue #160, opened 2019-08-02
ndt5: server should close download connection after 10sec


stephen-soltesz:
The legacy server used to close the download test connection immediately after it was done sending.

Large client populations are reported failures due to timeout because the new ndt-server waits to close the test connection until sending the TestFinalize message on the control connection.

For single-threaded clients, it is simpler to block on the test connection and then continue reading on the control connection.

stephen-soltesz:
Blocking https://github.com/m-lab/dev-tracker/issues/401

stephen-soltesz:
Resolved by https://github.com/m-lab/k8s-support/releases/tag/v1.2.5

bassosimone:
As reported in https://github.com/ooni/probe/issues/877#issuecomment-552773939, we noticed that the issue has resurfaced for OONI, so I'm opening this issue again.

With @robertodauria today we tested a fix that, deployed on sandbox, seems to fix OONI and all other clients that assume that the server closes the connection.

stephen-soltesz:
Thanks to @robertodauria & @bassosimone we tracked this down to an apparent accidental regression in closing all fds while collecting performance stats in https://github.com/m-lab/ndt-server/pull/179 -- @robertodauria has a fix in progress.

mehtab11:
@robertodauria - to confirm and then close.

PR #162, opened 2019-08-02
Explicitly close the download test connection after 10sec

stephen-soltesz:
Addresses https://github.com/m-lab/ndt-server/issues/160

As part of the fdcache, the server maintains two fds for the same client test connections. Both must be closed for the client to see the connection terminated.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/162)


stephen-soltesz:
https://github.com/m-lab/dev-tracker/issues/401

PR #165, opened 2019-08-05
Complete removal of close in MeasureViaPolling

stephen-soltesz:
This is a change that I failed to save so didn't get committed in https://github.com/m-lab/ndt-server/pull/162

Without this change, the server should still behave correctly but will attempt to close the fd twice, which should fail harmlessly.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/165)


bassosimone:
(lgtm)

PR #170, opened 2019-08-12
fix(ndt5/c2s): start upload when connection is ready

### LINKED ISSUE #89: ndt-browser-client.js:253 Uncaught DOMException: Failed to execute 'send' on 'WebSocket': Still in CONNECTING state.
nan

It's possible that a portion of the issue reported in https://github.com/m-lab/k8s-support/issues/127 is due to the ndt-browser-client.js failing to wait until the websocket connection is fully open before sending.

That issue may be addressed using an `.onopen` event, like:
https://github.com/ratchetphp/Ratchet/issues/481#issuecomment-273187668

```
var conn = new WebSocket('ws://localhost:8080/echo');
conn.onmessage = function(e){ console.log(e.data); };
conn.onopen = () => conn.send('hello');
```

bassosimone:
Closes #89

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/170)

### DIFF
--- a/html/ndt-browser-client.js
+++ b/html/ndt-browser-client.js
@@ -236,8 +236,7 @@ NDTjs.prototype.ndtC2sTest = function () {
     totalSent = 0,
     nextCallback = that.updateInterval,
     keepSendingData,
-    connectionOpen = false,
-    waitToStart;
+    connectionOpen = false;
 
   for (i = 0; i < dataToSend.length; i += 1) {
     // All the characters must be printable, and the printable range of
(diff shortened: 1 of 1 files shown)


bassosimone:
@ryanrolds:

> I looked over the code and it should solve the race condition, but I think it has the side effect of shifting some connection setup time into the test. After looking at the NDT server code, I believe delaying the sending of TEST_START until the client finished connecting to the test connection addresses the root cause of this race condition.

I believe the legacy implementation of ndt and this implementation are different and there is something the legacy implementation did that this one does not. In that sense, it is correct to see that the root cause of the issue lies in such change.

However, it is my understanding that the `onopen` event of a WebSocket is there precisely to know when a WebSocket is open and that JavaScript code is not supposed to call `.send` before `onopen` has fired.

I see this issue as fundamental, in the sense that the JavaScript code was working earlier either because we're lucky or because there was some fundamental assumption baked into the JavaScript code regarding the server behaviour that was not written in a comment.

Either way, making the code correct JavaScript code is, in my view, more correct that not doing it, regardless of what happens to the server side implementation.

I do also see this diff as ugly and I believe it would be less ugly with `async` and `await`. I have checked [the current state of async functions](https://caniuse.com/#feat=async-functions) and the main issue I see is that it won't work with Internet Explorer; do we care? If we don't care about Internet Explorer, then I can write a diff using that functionality.

stephen-soltesz:
I think I need to close this PR and open a new one so that I can assign @bassosimone as the reviewer.

PR #173, opened 2019-08-18
Adjusted run network to get NDT5 speed tests working

ryanrolds:
I wasn't able to get NDT5 tests working until I switched to using network=host as that exposes all ports bound by the container, not just the explicitly mapped/exposed ports. A side effect is that the service couldn't bind to 443 (possibly caused by the cap drop), updated to use 4443 instead.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/173)

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,8 +37,7 @@ sudo modprobe tcp_bbr
 and run the `ndt-server` binary container
 
 ```bash
-docker run --network=bridge                \
-           --publish 443:4443              \
+docker run --network=host                \
            --volume `pwd`/certs:/certs:ro  \
            --volume `pwd`/datadir:/datadir \
            --read-only                     \
@@ -53,7 +52,7 @@ docker run --network=bridge                \
 
 Once you have done that, you should have a ndt5 server running on ports
 `3001` (legacy binary flavour), `3002` (WebSocket flavour), and `3010`
-(secure WebSocket flavour); a ndt7 server running on port `443` (over TLS
+(secure WebSocket flavour); a ndt7 server running on port `4443` (over TLS
 and using the ndt7 WebSocket protocol); and Prometheus metrics available
 on port `9990`.
 
@@ -62,7 +61,7 @@ appear invalid to your browser, but everything is safe because this is a test
 deployment, hence you should ignore this warning and continue):
 
 * ndt5+wss: https://localhost:3010/static/widget.html
-* ndt7: https://localhost/static/ndt7.html
+* ndt7: https://localhost:4443/static/ndt7.html
 * prometheus: http://localhost:9090/metrics
 
 Replace `localhost` with the IP of the server to access them externally.


ryanrolds:
Why wouldn't you want to be able to run both at the same time? Having two different methods for running just complicates manually testing NDT7 and NDT5 quickly. The attempt of this PR is

bassosimone:
They are different use cases. The one which is there works for ndt7 with minimal privileges (and addresses my use case of testing ndt7 without caring about ndt5), while your use case is more tailored towards ndt5 and would break my test case in that I need traffic on 443 when I do research on a ndt7 feature before landing it on this repo.

I understand that I could use iptables to redirect the traffic, however, I am not happy about adding additional hoops to jump into before being up to speed.

Can you test whether adding —cap-add=NET_ADMIN and —network=host is enough to allow for a deployment that works for both ndt7 and ndt5? In such case, then maybe we can recommend that as a least privilege way of running the container.

ryanrolds:
In production how is it configured? Do both work?

ryanrolds:
I assume in production it's using 443 but I bet NDT5 also work, which is what I'm after. I guess I don't view the ports being different as a significant issue it's probably just a configuration option (flag or env var) on the client. What I view as a major issue is not being able to quickly run each speed tests. All developers are lazy (especially myself) and we will skip steps if it's not something we can quickly do. Ideally, a developer's local environment would startup with everything working.

Considering we already had an NDT5 breaking change released, one that I was able to reproduce (at least on 50% of tests) in my local env once I setup `ntd-server` and pointed my local SUA env at it, I think it would make sense to side on making it easier to run NDT5 speed tests easily.

bassosimone:
> In production how is it configured? Do both work?

I am not super familiar with how we do things in production. IIRC we are not actively working to reduce privileges, which was what I was after with this implementation here.

bassosimone:
@ryanrolds I am not familiar with Docker compose, so I don't understand what working with that tool entails. As far as this PR is concerned, I understand that providing more than one command to run the server is non practical. Also, do I understand correctly that after your change the code is able to run without issues and without privileges (it seems so, but I haven't tested it)? If that is true, then, do you mind listing an `iptables` command to make sure we redirect port `443` to `4443` such that the server is still exposed on `443`? I like the idea of having a single command to do all and you're right that that is more practical, but I'd also like to keep `443` working as before.

ryanrolds:
Docker Compose allows taking all the options/switches in the `docker run` and putting them in a single place (env vars are preferred, but switches work too). It also allows multiple services to t be defined in a single place. IMO, it's a must use and it really enables setting up microservices in Docker. Here is the SUA docker-compose.yml (https://github.com/Hack4Eugene/SpeedUpAmerica/blob/master/docker-compose.yaml). The SUA file isn't perfect, but it shows how it can be used. I have a mostly working docker-compose.yml working for ndt-server.

I will update the PR this evening. I may include the docker-compose.yml file I have mostly done.

ryanrolds:
I will test with `NET_ADMIN` too. That may make the need for iptables moot. I have no objections to binding to 443.

bassosimone:
Thanks!

ryanrolds:
Update. I'm still working on this. My evenings have been busy with a course that ends this week.

ryanrolds:
The main purpose of this PR was to make the project easier to set up for junior devs, but considering only Docker on Linux is supported that goal isn't possible. As much I think everyone should just use Linux, it's an unrealistic expectation. Closing the PR as it's not needed/useful.

PR #174, opened 2019-08-19
Add schema version to NDTResult struct

stephen-soltesz:
This change adds a new field for the "SchemaVersion" -- which should always be set to the "CurrentSchemaVersion" in the server -- so that archival data can continue to be parsed as the structure evolves over time. Using the schema version, ETL parser can be sure all available data is read correctly.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/174)

### DIFF
--- a/data/result.go
+++ b/data/result.go
@@ -26,10 +26,6 @@ type NDTResult struct {
 	GitShortCommit string
 	// Version is the symbolic version (if any) of the running server code.
 	Version string
-	// SchemaVersion represents the version of the NDTResult structure. This is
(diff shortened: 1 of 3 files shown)


pboothe:
Long in-person talk later:
Let's add smarts to the schema parser, and, in the ndt-server codebase, adopt the rule that a field can only be added and never deleted.  Then we don't need schema versioning, which represents a tarpit for parsers.

stephen-soltesz:
@pboothe -- updated the NDTResult comment PTAL?

PR #180, opened 2019-09-06
spec/ndt7: address several outstanding issues

bassosimone:
This modification of the ndt7 spec is non backward compatible
therefore I have bumped the version accordingly.

When reading the spec after this changes have been applied, I am now
confident that we can make a first public release soon.

The following list describes what changed:

1. bumped the version number

2. added a non-normative section describing the design choices of
   ndt7 as recently discussed with @pboothe and @mattmathis

3. reorganized the protocol description to group handshake-related
   information and websocket related information

4. [API CHANGE] now the client MUST include an user-agent
   during the websocket handshake

5. [API CHANGE] removed ambiguity on how the server SHOULD process
   query string keys without a value: now the server SHOULD consider
   them like keys with an empty string as value

6. [API CHANGE] do not force the server to use `400` to indicate
   handshake failure, be relaxed and use `4xx`

7. [API CHANGE] now the server MUST parse the query string

8. [API CHANGE] now the server MUST store the query string

9. reword the section where we describe usage of binary and textual
   websocket messages

10. [API CHANGE] say that implementations SHOULD follow the websocket
    RFC but concede that what they really MUST do is to be robust
    with respect to receiving errors once the test is started. This
    allows us to implement protections such that the duration of a
    test is precise as well as to allow servers to arbitrarily stop
    a transfer earlier than ten seconds if they need that.

11. [API CHANGE] switch to CamelCase for the measurement message

12. [API CHANGE] only document the minimum set of `TCP_INFO` variables
    that seem useful to OONI or Asurion and explicitly say that any
    other variable that MAY be there may disappear without notice

13. [API CHANGE] use natural kernel measurement units, e.g. bytes and
    microsecond, and do that consistently across the structure

14. [API CHANGE] clarify ConnectionInfo and add some extra optional
    fields the Go client was already using so that it will not conflict
    with future variables with same name and different semantics

15. add a section with an explicit example describing the messages
    exchanged both when there are and there are not measurement messages

16. clarify in mlab-ns section that we'll perform A/B testing and that
    implementing a client compatible with the specification is good
    precisely because it gives us flexibility in that respect

17. improve the example code describing exponential backoff to clearly
    indicate that exponential backoff is only to be done when mlabns
    returns the no-content signal

18. improve the reference implementations section to mention that
    simplified clients exist in this repository and to clarify that
    their main purpose is to test against this server

19. write an appendix section that is non normative and mention (1)
    the algorithm proposed by @pboothe for scaling the message size and
    (2) how the `TCP_INFO` variables included into this specification
    could be useful to people.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/180)

### DIFF
--- a/spec/data-format.md
+++ b/spec/data-format.md
@@ -57,8 +57,8 @@ Valid JSON metadata object in ClientMetadata could look like this:
 
 ```JSON
 {
-  "client_library_name":"libndt7.js",
-  "client_library_version":"0.4"
+  "ClientLibraryName": "libndt7.js",
+  "ClientLibraryVersion": "0.4"
 }
 ```
 
@@ -69,21 +69,16 @@ represent individual measurements recorded by the client or server.
 
 A measurement is a JSON object containing the fields specified by
 [ndt7-protocol.md](ndt7-protocol.md) in the "Measurements message" section,
-except that a server MAY choose to remove the "connection_info" optional
+except that a server MAY choose to remove the "ConnectionInfo" optional
 object to avoid storing duplicate information.
 
 A valid measurement JSON could be:
 
 ```JSON
 {
-  "bbr_info": {
-    "max_bandwidth": 12345,
-    "min_rtt": 123.4
-  },
-  "elapsed": 1.2345,
-  "tcp_info": {
-    "rtt_var": 123.4,
-    "smoothed_rtt": 567.8
+  "AppInfo": {
+    "ElapsedTime": 1234,
+    "NumBytes": 1234
   }
 }
 ```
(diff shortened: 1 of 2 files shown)


bassosimone:
@pboothe with respect dealing with unknown fields and setting them to `-1`, I think I have changed my mind. I've just had a chat with @mattmathis where he suggested that there is a very effective way of setting everything to `-1`, which is `memset(buf, 0xff, sizeof (buf))`. I didn't think about this and it's very neat and future proof. So, I think `-1` is really viable as a suggestion.

PR #188, opened 2019-09-23
Address three P0 ndt7 issues: PASTA, ping, JavaScript

bassosimone:
1. Use exponential function to select sleep time to get PASTA properties

2. Start sending pings to clients on an experimental basis (we will decide later how to use them and whether they are useful, we want them implemented now because we want under-development clients to handle - or ignore - pings gracefully)

3. Replace the JavaScript client with the one I used for research in August, which is more simple and should have all the features we need

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/188)


spec/ndt7-protocol.md
```diff
@@ -163,6 +163,15 @@ be implemented so that immediately after thirteen seconds have elapsed, the
 underlying TLS connection is closed. This can be implemented, e.g., in C/C++
 using alarm(3) to cause pending I/O operations to fail with `EINTR`.
 
+As regards ping and pong messages, a ndt7 server MAY send periodic
+ping messages during any test. Servers MUST NOT send ping messages more
+frequently than they would send textual messages. Clients SHOULD be prepared
+to receive ping messages. They SHOULD reply to such messages with pong
+messages containing the same payload. If a client is receiving too
+many ping messages in a specific time interval, it MAY drop them at
+random. A client MAY decide to ignore ping messages. Clients that
```
pboothe: Ignoring PING is contrary to the websocket spec.  Let's not allow NDT7 clients to behave in ways which contradict the websocket RFC.

pboothe: https://tools.ietf.org/html/rfc6455#section-5.5.2 -    Upon receipt of a Ping frame, an endpoint MUST send a Pong frame in response, unless it already received a Close frame.  It SHOULD respond with Pong frame as soon as is practical.

ndt7/spec/spec.go
```diff
@@ -28,8 +28,19 @@ const MaxScaledMessageSize = 1 << 20
...
-// MinMeasurementInterval is the minimum interval between measurements.
-const MinMeasurementInterval = 250 * time.Millisecond
+// AveragePoissonSamplingInterval is the average of a lambda distribution
+// used to decide when to perform next measurement.
+const AveragePoissonSamplingInterval = 250 * time.Millisecond
+
+// MinPoissonSamplingInterval is the min acceptable time that we want
+// the lambda distribution to return. Smaller values will be clamped
+// to be this value instead.
+const MinPoissonSamplingInterval = 25 * time.Millisecond
```
pboothe: You don't have to set min and max. Why are you choosing to do so here?

bassosimone: I'm more worried by not choosing max than by not choosing min. The lower bound is set because we don't want to send too close messages. The upper bound because we want ~1 message per second to update the UI.

spec/ndt7-protocol.md
bassosimone: That text was meant to make half-duplex clients compliant during the upload. I am amending the text to omit that text. I believe it's reasonable to argue that, for an half-duplex client, as soon as practical means after it has finished uploading ;-). Do you want me to explicit that?

bassosimone: See 0dc5a4f

ndt7/spec/spec.go
bassosimone: (Incidentally, those limits are the ones recommended by the docs. I can inflate the upper bound as much as one second, if you wish. But I'd like to try to generate ~1 message per second in the worst case. Both for the client's sake and for the data saving sake [*]).

[*] What is the frequency of sampling of the sidecar service? Do we want PASTA also there?

pboothe: Thanks!  All I wanted was a reason. "Because 1 per second is required for good UX" is a great reason.

the tcp_info sidecar service samples at max rate. Whether it should be poisson instead is an interesting question.

spec/ndt7-protocol.md
pboothe: New language looks good to me.

bassosimone:
Thanks, merging!

PR #209, opened 2019-10-21
Connection ID fallback and Docker Compose support

ryanrolds:
I put the ID fallback in `fdcache` because the `m-lab/uuid` appears to function as a wrapper for getting an ID via the TCP connections SO_COOKIE and I don't think it makes sense to muddy that package with fallback support.

The PR also adds support for Docker Compose. The `Dockerfile.local` is used by `docker-compose.yml` because the main `Dockerfile` doesn't copy `gen_local_test_certs.sh` into the image. I also switch `gen_local_test_certs.sh` to use /bin/sh as I'm pretty sure Alpine Linux doesn't include Bash. I also added the cert moving into the script.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/209)


gen_local_test_certs.sh
```diff
@@ -1,6 +1,7 @@
-#!/bin/bash
+#!/bin/sh
```
bassosimone: Can you test whether this works with dash and ksh? I believe the main reason why it was named bash and it was invoking explicitly bash is because it is using the non-portable `set -euxo pipefail`.

ryanrolds: You're correct, it doesn't work under Dash. I'm switching the interpreter back to Bash and updating `Dockerfile.local` to install Bash.

bassosimone:
Will merge when Travis becomes green

PR #212, opened 2019-10-29
attempted fix for #211

### LINKED ISSUE #211: Full stack dockerfile start.sh script should check for prefix file before trying to create it
When starting a fullstack container, the script [start.sh](https://github.com/m-lab/ndt-server/blob/master/fullstack/start.sh#L37) writes a `prefix` file each time the container is started. This is fine for demonstration, but not great for operational use, since we might start and stop a container and want to have persistent storage of test results across those restarts. Currently when you start this container a second time using a persistent storage volume, the script fails on linux systems (despite the `-p` flag), requiring the operator to manually remove the `prefix` file.

Adding a basic check that only tries to create the prefix file if it doesn't exist should solve this.

critzo:
This is an untested draft PR attempting to fix #211 @pboothe PTAL?

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/212)

### DIFF
--- a/fullstack/start.sh
+++ b/fullstack/start.sh
@@ -30,7 +30,9 @@ set -euxo pipefail
 UUID_DIR=/var/local/uuid
 UUID_FILE=${UUID_DIR}/prefix
 mkdir -p "${UUID_DIR}"
-/create-uuid-prefix-file --filename="${UUID_FILE}"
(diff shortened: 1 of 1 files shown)


pboothe:
LGTM

PR #213, opened 2019-10-30
UUIDs are now in tmpfiles

pboothe:
After some thought last night, I realized we moved the codebase in the wrong direction.  Instead, the UUID prefix should be in a file in `/tmp` and be recreated on every run.  Keeping the old file around puts the uniqueness of the UUIDs at risk.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/213)


pboothe:
Thanks! Let's merge this!

PR #179, opened 2019-09-03
Smooth c2s

pboothe:
Eliminates quantization of c2s upload results, adds unit tests, and eliminates several potential sources of goroutine leaks.

Part of https://github.com/m-lab/dev-tracker/issues/412
Fixes #144 

This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/179)

PR #218, opened 2019-11-13
Fix timeout at the end of S2C tests

robertodauria:
This PR adds a `defer fp.Close()` in the `measureUntilContextCancellation` func so that the connection gets closed when the context is cancelled. At some point in the past that `Close()` was removed (https://github.com/m-lab/ndt-server/pull/179), it's unclear to me why.
 
This should fix the timeout some clients are intermittently experiencing at the end of a S2C test. While some clients are robust enough to ignore the timeout, others are currently failing.

Hopefully fixes https://github.com/ooni/probe/issues/877 once released.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/218)


bassosimone:
@robertodauria can you add a comment explaining why the close matters? I guess this will reduce the likelihood that it gets removed in the future, maybe?

Issue #228, opened 2020-01-09
fullstack image is busted


pboothe:
discovered by @laiyi-ohlsen

pboothe:
Fixed in #229

PR #229, opened 2020-01-09
Fix the way we are calling the fullstack components

pboothe:
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/229)


pboothe:
Fixes #228

pboothe:
Pushing without passing tests due to breakage in libndt that is not our fault.

PR #231, opened 2020-01-10
Add concurrent connection limits to NDT server

stephen-soltesz:
This change adds a new `access.MaxController` that when enabled, will allow up to `Max` connections. Default is disabled.

This simple connection limiter is applied to all service ports: 3001 (ws/raw), 3010 (wss), and 443 (ndt7). And, the connection limiter is not applied to the metrics server. Meaning that even under load, the server should remain monitorable.

In time, the "Controller" interface can be extended to support access_tokens, and individual client rate limiting.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/231)

### DIFF
--- a/TestDockerfile
+++ b/TestDockerfile
@@ -11,7 +11,7 @@
 
 
 # A base image for building and the final image.
-FROM golang:1.12-buster AS ndtbase
+FROM golang:1.13-buster AS ndtbase
 WORKDIR /
 RUN apt-get update && apt-get install -y git libmaxminddb0 libevent-2.1-6 \
     libevent-core-2.1-6 libevent-extra-2.1-6 \
--- a/access/controller.go
+++ b/access/controller.go
@@ -1,0 +1,28 @@
+package access
+
+import (
+	"net/http"
+	"sync/atomic"
+)
+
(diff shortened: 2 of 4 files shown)


stephen-soltesz:
I don't think we'll use this strategy.

mattdlh:
> I don't think we'll use this strategy.

@stephen-soltesz was another strategy for managing concurrent connections to the server implemented?

pboothe:
Yes. Requiring time-limited secure access tokens handed out by https://github.com/m-lab/locate , our centralized geography-aware and load-aware load-balancer.  The NDT server stays dumb, only the load balancer has to be smart.

stephen-soltesz:
I can also add that the "tx controller" logic is evaluated by the NDT server.

See options:

```
  -txcontroller.device string
  -txcontroller.max-rate uint
```

The tx controller allows the server to specify an out-bound network utilization "max-rate". When the server observes utilization above this rate, no new connections are allowed. For example, a server on a 1Gbps link that wants to service clients that also have 1Gbps capacity should set the -txcontroller.max-rate quite low, maybe 100Mbps. This approach is more dynamic and responsive to constraints where a fixed connection limit (this PR) cannot.

Issue #192, opened 2019-09-25
ndt7: save application level RTT samples


bassosimone:
In #188, we have implemented the mechanism for measuring application-level pings RTT but we are not currently saving it. This should be done, to make the data actionable. We need to update the spec to mention this piece of data and change the code to save it.

bassosimone:
BTW, issue #201 seems to be a case where the RTT is actually 20-ish but BBR thinks it is much smaller, therefore, it would be beneficial to see what applevel-RTT says about the connection.

darkk:
I've done some shallow refactoring to include base `time.Time` reference, save websocket-ping RTT and send it to client. Unfortunately, measuring L7 RTT while doing `/upload` or `/download` produces significantly skewed results those are hundreds milliseconds apart. That's expected at least due to TCP HOL, so I've added `/ping` endpoint to measure RTT cleanly.

The test produces expected difference between L7 and L4 RTT when TCP proxy is present on path. E.g. Tor Browser measures L7 / L4 RTT as `302.6 / 5.9 ms` from my location in St. Petersburg to a [VM in London](https://fish.darkk.net.ru:44433/static/ndt7.html).

It's also theoretically possible to modify `/download` spec to collect L7 RTT samples before pushing the actual data to avoid HOL. It's unclear to me how it may affect existing clients as I'm not that aware of the client logic that is already rolled out.

What do you think, should it be `/ping` or should `/download` spec rather be amended?

bassosimone:
@darkk considering how other speed tests work, I think may be preferable to include an application level `/ping` phase to the experiment. I'm significantly less happy about modifying the expected behavior of the client during `/download` or `/upload` because that would complicate the implementation and we wanted to keep the client's implementation very easy.

PR #234, opened 2020-01-21
Saving websocket RTT samples

darkk:
That's WIP for #192, reopened per [the comment](https://github.com/m-lab/ndt-server/pull/232#issuecomment-576624672) in #232.

I have a question I'm unsure about:

- is it okay to carry start around the way it's implemented? It looks a bit messy to me, but I'm quite okay with that.

WIP is:
- [ ] send only the first `WSInfo` during _download_ test. It breaks the feedback loop and makes queue management easier. Biased RTT estimates are not so interesting in the real time.

### DIFF
--- a/ndt7/download/download.go
+++ b/ndt7/download/download.go
@@ -3,6 +3,7 @@ package download
 
 import (
 	"context"
+	"time"
 
 	"github.com/gorilla/websocket"
 	"github.com/m-lab/ndt-server/ndt7/download/sender"
@@ -15,13 +16,15 @@ import (
 // Do implements the download subtest. The ctx argument is the parent
 // context for the subtest. The conn argument is the open WebSocket
 // connection. The resultfp argument is the file where to save results. Both
-// arguments are owned by the caller of this function.
-func Do(ctx context.Context, conn *websocket.Conn, resultfp *results.File) {
+// arguments are owned by the caller of this function. The start argument is
+// the test start time used to calculate ElapsedTime and deadlines.
+func Do(ctx context.Context, conn *websocket.Conn, resultfp *results.File, start time.Time) {
 	// Implementation note: use child context so that, if we cannot save the
 	// results in the loop below, we terminate the goroutines early
 	wholectx, cancel := context.WithCancel(ctx)
 	defer cancel()
-	senderch := sender.Start(conn, measurer.Start(wholectx, conn, resultfp.Data.UUID))
-	receiverch := receiver.StartDownloadReceiver(wholectx, conn)
+	measurerch := measurer.Start(wholectx, conn, resultfp.Data.UUID, start)
+	receiverch, pongch := receiver.StartDownloadReceiver(wholectx, conn, start, measurerch)
+	senderch := sender.Start(conn, measurerch, start, pongch)
 	saver.SaveAll(resultfp, senderch, receiverch)
 }
--- a/ndt7/model/measurement.go
+++ b/ndt7/model/measurement.go
@@ -8,4 +8,5 @@ type Measurement struct {
 	ConnectionInfo *ConnectionInfo `json:",omitempty" bigquery:"-"`
 	BBRInfo        *BBRInfo        `json:",omitempty"`
 	TCPInfo        *TCPInfo        `json:",omitempty"`
+	WSInfo         *WSInfo	       `json:",omitempty"`
 }
--- a/ndt7/receiver/receiver.go
+++ b/ndt7/receiver/receiver.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"encoding/json"
 	"time"
+	"math"
 
 	"github.com/gorilla/websocket"
 	"github.com/m-lab/ndt-server/logging"
@@ -19,28 +20,44 @@ type receiverKind int
 const (
 	downloadReceiver = receiverKind(iota)
 	uploadReceiver
+	pingReceiver
+)
+
+const (
+	MaxDuration = math.MaxInt64 * time.Nanosecond
 )
 
 func loop(
 	ctx context.Context, conn *websocket.Conn, kind receiverKind,
-	dst chan<- model.Measurement,
+	dst chan<- model.Measurement, start time.Time, pongch chan<- model.WSInfo,
 ) {
 	logging.Logger.Debug("receiver: start")
 	defer logging.Logger.Debug("receiver: stop")
 	defer close(dst)
+	defer close(pongch)
 	conn.SetReadLimit(spec.MaxMessageSize)
 	receiverctx, cancel := context.WithTimeout(ctx, spec.MaxRuntime)
 	defer cancel()
-	err := conn.SetReadDeadline(time.Now().Add(spec.MaxRuntime)) // Liveness!
+	err := conn.SetReadDeadline(start.Add(spec.MaxRuntime)) // Liveness!
 	if err != nil {
 		logging.Logger.WithError(err).Warn("receiver: conn.SetReadDeadline failed")
 		return
 	}
+	minRTT := MaxDuration
 	conn.SetPongHandler(func(s string) error {
-		rtt, err := ping.ParseTicks(s)
+		elapsed, rtt, err := ping.ParseTicks(s, start)
 		if err == nil {
-			rtt /= int64(time.Millisecond)
-			logging.Logger.Debugf("receiver: ApplicationLevel RTT: %d ms", rtt)
+			logging.Logger.Debugf("receiver: ApplicationLevel RTT: %d ms", int64(rtt / time.Millisecond))
+			if rtt < minRTT {
+				minRTT = rtt
+			}
+
+			wsinfo := model.WSInfo{
+				ElapsedTime: int64(elapsed / time.Microsecond),
+				LastRTT: int64(rtt / time.Microsecond),
+				MinRTT: int64(minRTT / time.Microsecond),
+			}
+			pongch <- wsinfo // Liveness: buffered (sender)
 		}
 		return err
 	})
@@ -55,11 +72,11 @@ func loop(
 		}
 		if mtype != websocket.TextMessage {
 			switch kind {
-			case downloadReceiver:
+			case uploadReceiver:
+				continue // No further processing required
+			default: // downloadReceiver and pingReceiver
 				logging.Logger.Warn("receiver: got non-Text message")
 				return // Unexpected message type
-			default:
-				continue // No further processing required
 			}
 		}
 		var measurement model.Measurement
@@ -72,10 +89,15 @@ func loop(
 	}
 }
 
-func start(ctx context.Context, conn *websocket.Conn, kind receiverKind) <-chan model.Measurement {
+func startReceiver(ctx context.Context, conn *websocket.Conn, kind receiverKind, start time.Time) (<-chan model.Measurement, <-chan model.WSInfo) {
+	// |dst| is going to the log file
 	dst := make(chan model.Measurement)
-	go loop(ctx, conn, kind, dst)
-	return dst
+	// |pongch| goes to the client, it's buffered to avoid blocking on `download.sender.loop`
+	// while `conn.WritePreparedMessage()` is active.
+	// TODO(darkk): is it possible to reduce buffer size or to avoiding blocking in some other way? May avoiding L7 pings at /download altogether be the way?
+	pongch := make(chan model.WSInfo, 1 + spec.MaxRuntime / spec.MinPoissonSamplingInterval)
+	go loop(ctx, conn, kind, dst, start, pongch)
+	return dst, pongch
 }
 
 // StartDownloadReceiver starts the receiver in a background goroutine and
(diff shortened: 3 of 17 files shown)


ndt7/download/download.go
```diff
@@ -15,13 +16,15 @@ import (
...
+func Do(ctx context.Context, conn *websocket.Conn, resultfp *results.File, start time.Time) {
 	// Implementation note: use child context so that, if we cannot save the
 	// results in the loop below, we terminate the goroutines early
 	wholectx, cancel := context.WithCancel(ctx)
 	defer cancel()
-	senderch := sender.Start(conn, measurer.Start(wholectx, conn, resultfp.Data.UUID))
-	receiverch := receiver.StartDownloadReceiver(wholectx, conn)
+	measurerch := measurer.Start(wholectx, conn, resultfp.Data.UUID, start)
+	receiverch, pongch := receiver.StartDownloadReceiver(wholectx, conn, start, measurerch)
+	senderch := sender.Start(conn, measurerch, start, pongch)
```
bassosimone: I am confused by the fact that a new channel has been introduced. Isn't it possible to use the `measurerch` for passing around information? I understood that the PING is another `null`able pointer within a Measurement.

ndt7/receiver/receiver.go
```diff
@@ -19,28 +20,44 @@ type receiverKind int
...
+			if rtt < minRTT {
+				minRTT = rtt
+			}
+
+			wsinfo := model.WSInfo{
+				ElapsedTime: int64(elapsed / time.Microsecond),
+				LastRTT: int64(rtt / time.Microsecond),
+				MinRTT: int64(minRTT / time.Microsecond),
+			}
+			pongch <- wsinfo // Liveness: buffered (sender)
```
bassosimone: So, here I believe it should be possible to use `dst` to emit a measurement containing PingInfo, right?

ndt7/download/download.go
darkk: `download.sender.loop` goroutine may be naturally blocked on `conn.WritePreparedMessage`, so there is a risk that `send(measurerch)` will block `receiver.loop` goroutine. Blocked `measurer.loop` is good, it'll just skip a few ticks. Blocked `receiver.loop` leads to skewed RTT measurements and I'd rather avoid that.

There are several possible options to overcome that.

Current implementation creates buffered channel for RTT measurements with a buffer size equal of maximum possible amount of pings issued during the test. It was good enough for PoC, but I consider it a wasteful behavior optimized for the worst case.

I'm refactoring the code to send less `WSPingInfo` messages to the client. Download sub-test will send only the first message back to the client via a buffered channel with small buffer and tail-drop other messages if the channel is not writable due to sender being blocked. The `WSPingInfo` messages would be still logged, and I assume that writing to the logging channel does not naturally block for an unpredictable amount of time (unlike `conn.WritePreparedMessage` in `download.sender.loop`). I think, logging `WSPingInfo` in `ClientMeasurements` section is okay. Anyway, the `elapsed` value in ping/pong frame is not signed, so client can fake it, so putting that to `ClientMeasurements` makes sense to me :-)

BTW, should ping be signed or formatted as `{ndt7: $value}` as a safety net against a client sending unsolicited pongs with some numerical value? Clients can send unsolicited pongs for heartbeat purposes [per RFC](https://tools.ietf.org/html/rfc6455#section-5.5.3).

It may make sense to apply the same logic to the measurements coming from `measurer.loop`. Currently the messages are logged if and only if `dst` is available for writing (sender is not blocked). Maybe it makes sense to split that channel into two: one channel coming to the log (that blocks on "disk") and another coming to the client (that may tail-drop messages if the client is not fast enough to fetch `WritePreparedMessage` blob). What do you think?

ndt7/model/measurement.go
```diff
@@ -8,4 +8,5 @@ type Measurement struct {
 	ConnectionInfo *ConnectionInfo `json:",omitempty" bigquery:"-"`
 	BBRInfo        *BBRInfo        `json:",omitempty"`
 	TCPInfo        *TCPInfo        `json:",omitempty"`
+	WSInfo         *WSInfo	       `json:",omitempty"`
```
darkk: @bassosimone, you've said at #232 that it should rather be `PingInfo`. I'd suggest `WSPingInfo` then to avoid possible confusion about different pings at different network layers. Does it sound okay?

Websocket ping-pong may be not end-to-end as well in some weird hypothetical cases involving frame-based websocket load balancer. I consider those cases extremely unlikely for `wss` but `ws` has some chances of alike network interference.

ndt7/download/download.go
darkk: BTW, slow client blocking TCPInfo logging seems to be a real-world case, not a hypothetical situation. I've checked my few ndt7 data samples, some of them have a gap as large as 1.2s between TCPInfo.ElapsedTime samples (that's twice as large as `spec.MaxPoissonSamplingInterval` being 625ms). Those 11 files with 321 TCPInfo samples have 15 TCPInfo samples being more than 650ms apart.

ndt7/model/measurement.go
bassosimone: Yes, the name you propose sounds better

ndt7/download/download.go
bassosimone: Are you talking about client side traces or server side traces?

bassosimone: That you don't get measurements on the client side because of head of line blocking is a know issue that I think we cannot address. However, the channel has buffer, so the measurements were actually saved on disk when I checked. The improvement that I foresee here is actually separating saving measurements from sending measurements to client. In the second case, I think it is reasonable to consider sending the client the most recent sample not the most ancient.

bassosimone: Another thought that I have is that I'd try to not change the channel structure in this PR but rather consider changing the channel structure later. The code is using a pipeline pattern and channels are closed to signal EOF. However, there are better patterns where there are multiple workers per stage and where contexts are used rather than closing channels. I have a sense that the right solution here is to measure and reorganise channels to have more parallelism when needed.

darkk: > However, the channel has buffer

As far as I see, the created channels do not have buffers, all `make(chan...)` calls under `ndt7/` do not specify any non-zero capacity. Do you mean golang channels here?

I was worried about the following backpressure situation:
- the client is _slow_ and his TCP-proxy has large TCP buffer (the easiest way for me to reproduce the case is `tor`)
- `download.sender.loop` goroutine blocks on `conn.WritePreparedMessage` for some time (`tor` case gives me delays like 8500ms)
- `download.sender.loop` does not fetch messages from `m, ok := <-src` as the goroutine is blocked
- `measurer.loop` goroutine blocks on sending to `dst` as the channel is not drained and has no buffer
- `measurer.loop` samples TCP_INFO and BRR_INFO possibly less often that `MaxPoissonSamplingInterval` as it's blocked on sending to `dst` and skips some of `ticker.C` ticks
- the TCPInfo samples are not written to the log as they're not actually sampled :-)

Note, I assume that `ndt7` has a (weak) guarantee to take TCPInfo samples at least as often as `MaxPoissonSamplingInterval`. That's just my assumption and it may be wrong.

This backpressure case worries me as it means that `receiver.loop` being blocked on writing RTT estimate to the channel drained by `sender` may provide wrong RTT estimate for the next sample in a queue. `receiver` should be mostly blocked on `conn.ReadMessage` to provide reasonably accurate data.

The orignal goal to introduce `pongch` was to mitigate that backpressure with quite a deep buffer. Bug I agree that this backpressure issue should be out of the scope of this pull request.

Back to answering your original question. My goal is go have some of `WSPingInfo` samples being sent back to the ndt7 client during the ping test runtime. The `master` branch has the channel passing both the messages and the EOF signal from `measurer.loop` to `${testname}.sender.loop`. Having two concurrent writers (`measurer.loop` and `receiver.loop`) sending to the channel that is eventually closed(!) will lead to goroutine panic (on an attempt to send to closed channel).

So, I'm still going to use `pongch` as a channel from `receiver.loop` to `${testname}.sender.loop` to keep the "pipeline + EOF" pattern.

ndt7/receiver/receiver.go
darkk: Yes, it may be written to the resulting json file here, it will be put under `ClientMeasurements` in this case. But `pongch` is still useful to send the message back to ndt7 client for the liveness reasons mentioned above.

ndt7/download/download.go
bassosimone: TL;DR I acknowledge the issue you report (did see it myself). I will not accept a patch with pongch. I explain how I’d change the architecture to avoid losing measurements. I am not asking you to implement that, however. I am just asking you to kindly focus here on a simple patch that implements the /ping you need and does not intersect with existing subtests.

> The channel has buffer [...]

I was wrong. I used a buffer in a development branch then. I must have concluded that was not the right fix, which is to change the architecture.

> [...] the pipeline+EOF pattern

That pattern is wrong. A pattern where an expired ctx tells you a goroutine to stop is right. We’ll fix that.

> [...] I am still using to use pongch 

Your attitude is wrong. The right attitude is to listen. We are on the same page on more than 99% of the topics here.

Let me summarize my position: thank your for exposing a bunch of architectural issues. I am not going to accept a patch that complicates significantly the code base. Please focus on writing a /ping endpoint that does its job and does not intersect with the implementation of other experiments.

That patch, I will expedite its review. I am not willing to spend further time to argue on the merits of adding complexity to the code base when this can be avoided. That is 無駄無駄無駄無駄.

> Worried about the following back pressure situation [...]

Your analysis is right. Your remediation is wrong. Thank your for spelling out this issue so clearly. As I mentioned to you (I guess I’m private?) I did see the same for a 2G connection.

> [...] weak guarantee to take TCP samples

This is right or wrong depending on the meaning of take. It’s right if take also implies save. It’s wrong if take also imply sending the data to the client. We send data to the client on a best effort basis but we should be doing our best to collect and save samples.

(On this note: M-Lab has a sidecar service that samples TCPInfo at high frequency. The data is collected separately and can be joined after submission. We are still collecting BBR and TCPInfo here, why? The original design was to stop early when BBR bandwidth estimation did become stable, but we have not implemented this yet. Anyway, this should help you understand why dealing with this issue has priority medium and not high.)

The right fix that significantly mitigates all that have been said so far is to change the code architecture as follows:

1. the measurer is a class and stores measurements in itself in a lock free way 

2. the channels become all non blocking when writing  

3. LIFO is better than FIFO when passing measurements downstream

4. we use a WaitGroup to wait for all the goroutines to complete

5. then we’re again single threaded and we just save on disk accessing directly what was measured

This is the right architectural fix for the current situation.

> goal is to send back samples to the client

We already agreed that your use case is best served by a separate subtest called ping. We already agreed that we also need a single sample at the beginning of the /download. Everything else seems irrelevant to the objective, hence wrong.

If adding support for WebSocket level ping to download and upload is such a burden with the current architecture, the right thing to do is to open an issue, just take the first sample of the connection, and get rid of the remaining code. A future code architecture may accommodate this feature with less effort. This is something that would tell us that the architecture is right. Now it is clearly wrong because it does not allow us to do that easily.

Instead, please focus on writing /ping to fulfill your goals. Please do so without complicating the implementation of existing endpoints in exchange for seriously marginal gains.

✌️

bassosimone: TL;DR I acknowledge the issue you report (did see it myself). I will not accept a patch with pongch. I explain how I’d change the architecture to avoid losing measurements. I am not asking you to implement that, however. I am just asking you to kindly focus here on a simple patch that implements the /ping you need and does not intersect with existing subtests.

> The channel has buffer [...]

I was wrong. I used a buffer in a development branch then. I must have concluded that was not the right fix, which is to change the architecture.

> [...] the pipeline+EOF pattern

That pattern is wrong. A pattern where an expired ctx tells you a goroutine to stop is right. We’ll fix that.

> [...] I am still using to use pongch 

Your attitude is wrong. The right attitude is to listen. We are on the same page on more than 99% of the topics here.

Let me summarize my position: thank your for exposing a bunch of architectural issues. I am not going to accept a patch that complicates significantly the code base. Please focus on writing a /ping endpoint that does its job and does not intersect with the implementation of other experiments.

That patch, I will expedite its review. I am not willing to spend further time to argue on the merits of adding complexity to the code base when this can be avoided. That is 無駄無駄無駄無駄.

> Worried about the following back pressure situation [...]

Your analysis is right. Your remediation is wrong. Thank your for spelling out this issue so clearly. As I mentioned to you (I guess I’m private?) I did see the same for a 2G connection.

> [...] weak guarantee to take TCP samples

This is right or wrong depending on the meaning of take. It’s right if take also implies save. It’s wrong if take also imply sending the data to the client. We send data to the client on a best effort basis but we should be doing our best to collect and save samples.

(On this note: M-Lab has a sidecar service that samples TCPInfo at high frequency. The data is collected separately and can be joined after submission. We are still collecting BBR and TCPInfo here, why? The original design was to stop early when BBR bandwidth estimation did become stable, but we have not implemented this yet. Anyway, this should help you understand why dealing with this issue has priority medium and not high.)

The right fix that significantly mitigates all that have been said so far is to change the code architecture as follows:

1. the measurer is a class and stores measurements in itself in a lock free way 

2. the channels become all non blocking when writing  

3. LIFO is better than FIFO when passing measurements downstream

4. we use a WaitGroup to wait for all the goroutines to complete

5. then we’re again single threaded and we just save on disk accessing directly what was measured

This is the right architectural fix for the current situation.

> goal is to send back samples to the client

We already agreed that your use case is best served by a separate subtest called ping. We already agreed that we also need a single sample at the beginning of the /download. Everything else seems irrelevant to the objective, hence wrong.

If adding support for WebSocket level ping to download and upload is such a burden with the current architecture, the right thing to do is to open an issue, just take the first sample of the connection, and get rid of the remaining code. A future code architecture may accommodate this feature with less effort. This is something that would tell us that the architecture is right. Now it is clearly wrong because it does not allow us to do that easily.

Instead, please focus on writing /ping to fulfill your goals. Please do so without complicating the implementation of existing endpoints in exchange for seriously marginal gains.

✌️

darkk: _(writing that down, so the essence is not lost in a log of a private chat)_

I've refactored the code under following rules:
- keep `measurer` and `saver` intact
- avoid two ping frames in-flight (drop ping sample if the previous pong has not arrived yet)
- have LIFO logic for samples sent to client
- make the channels "almost non-blocking" for sender. I mean structuring the code in a way that blocking on `chan <- value` MAY happen in some case, but this case is an "erratic" code path, not a "usual" one.
- use EOF as a completion signal for most of the goroutines to terminate them in a predictable order
- use context.cancel() to stop memoryless timer early in case of a network error from a client

> complicating ... in exchange for seriously **marginal** gains

Yeah, I'm sorry, that's an (unlucky) trait I have indeed! Thanks (no kidding) for nudging me towards awareness, better focus and simplicity.

> M-Lab has a sidecar service that samples TCPInfo at high frequency

I've also completely forgotten about that. That's why I overestimated the utility of TCPInfo samples being logged.

PR #250, opened 2020-02-06
Make ndt7 upload client simpler

pboothe:
Simpler means fewer places for bugs to hide.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/250)


pboothe:
Merging without waiting for tests to pass - only JS changed and the JS is currently untested, so no test status can be changed.

PR #252, opened 2020-02-10
ndt7: start improving ping implementation

### LINKED ISSUE #242: Merge simple changes from #234 to master
This basically amounts to:

1. fixes that propagate the current start time correctly

2. fixes that make sure we send a single RTT sample

It should also be an interesting follow-up to consider how refactoring the channels as discussed in #238 can be helpful to actually allow to save application-level measurement samples.

bassosimone:
This diff has been written by @darkk as part of #234 and
has been extracted #234 by @bassosimone, with the objective
of merging these low hanging fruits while we look into the
proper way of implementing the `/ping` endpoint.

Closes #242


bassosimone:
Oh, noes, conflicts! I am going to solve them now!

bassosimone:
@pboothe I have fixed conflicts, can you perhaps maybe take another look?

gfr10598:
FYI, I'd very much like to see a ping beacon client, that would ping many nearby servers, and a random set of far away servers, every 5 minutes or so.  But we need a server that would collect the data, too.

PR #253, opened 2020-02-10
Remove connection to ETL pipeline.

pboothe:
Now that ETL is using modules, this bidirectional "code health" check has strictly negative value for NDT-server.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/253)

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -6,10 +6,6 @@ set -ex
 # Test the NDT binary
 PATH=${PATH}:${GOPATH}/bin
 
-# Check that we haven't blatently broken etl dependencies.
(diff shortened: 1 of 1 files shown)


pboothe:
In particular, etl is broken at head. I'd like to get that fixed, but I'd like to, even more, make it so that broken ETL does not cause a broken NDT server.

pboothe:
Because it is frustrating to have the build success for ndt-server be the floor of the build success for all of its dependencies AND all of the things for which ndt-server is a dependency.  We can't escape the first, but I would like to try and avoid the second.

PR #259, opened 2020-02-12
Add access control to ndt7:80 and ignore TxController errors

stephen-soltesz:
This change adds the access controller to ndt7 clear text handler and generalizes the access controller setup to allow chaining multiple controllers in the future. This change also allows TxController to be disabled with a warning.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/259)

### DIFF
--- a/ndt-server.go
+++ b/ndt-server.go
@@ -123,6 +123,8 @@ func main() {
 		// Only watch interface and run tx.Limit on success.
 		go tx.Watch(ctx)
 		ac.Append(tx.Limit)
+	} else {
+		log.Println("WARNING: access.TxController disabled:", err)
 	}
 
 	// The ndt5 protocol serving non-HTTP-based tests - forwards to Ws-based
--- a/ndt-server.go
+++ b/ndt-server.go
@@ -10,11 +10,9 @@ import (
 	"os/signal"
 	"syscall"
 
-	"github.com/m-lab/go/prometheusx"
-
 	"github.com/m-lab/go/flagx"
+	"github.com/m-lab/go/prometheusx"
 	"github.com/m-lab/go/rtx"
-
 	"github.com/m-lab/ndt-server/access"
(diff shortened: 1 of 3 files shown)


ndt-server.go
```diff
@@ -24,21 +22,22 @@ import (
 	"github.com/m-lab/ndt-server/ndt7/spec"
 	"github.com/m-lab/ndt-server/platformx"
 
+	"github.com/justinas/alice"
```
bassosimone: Is this middleware chaining library named alice because of Alice in Chains?

ndt-server.go
```diff
@@ -117,9 +116,16 @@ func main() {
...
+	ac := alice.New()
 	tx, err := access.NewTxController(*maxRate)
-	rtx.Must(err, "Failed to allocate new TxController")
-	go tx.Watch(ctx)
+	if err == nil {
+		// Only watch interface and run tx.Limit on success.
+		go tx.Watch(ctx)
+		ac.Append(tx.Limit)
+	} else {
+		log.Println("WARNING: access.TxController disabled:", err)
```
bassosimone: I am confused by why the access controller, when disabled, returns an error? Perhaps I just don't know enough about how it works, but I'd have found more understandable if it returned something like `tx.ErrDisabled`.

(Is there a scenario where we want to disable it? Maybe is that testing? Can we maybe use command line for that?)

ndt-server.go
stephen-soltesz: Hahaha, you must be right. I originally thought it was Alice in Wonderland, but the package literally has  `alice.Chains` data type.

ndt-server.go
stephen-soltesz: We have two flags currently for this feature, -txcontroller.device and -max-rate.

I've changed the default value for device to the empty string, requiring that it be specified to be enabled. What do you think?

Issue #260, opened 2020-02-12
Separate data.NDT5Result and data.NDT7Result structs


stephen-soltesz:
This has several implications for m-lab/etl and m-lab/website.

In particular, the current code uses a shared/unified `data.NDTResult` struct. The etl/schema imports that struct directly and uses it to generate a schema.

Unfortunately, recent changes to the ndt7 data types have made the data.NDTResult struct incompatible with the current BQ schema.

To solve this, the proposal is to create two separate data.NDTResult structs (one for ndt5 & ndt7) with protocol-specific fields.

This should remain backward compatible with archived data for both tests (with the exception of the fields removed from ndt7). And, this will make the schemas generated for the respective datatypes type-specific (which is good) and the documentation for those tables include only relevant fields for that data type (also good).

Our thinking about how to create "unified tables" with this data has evolved to include "Standard Columns" and the Unified NDT history view. Going forward, we do not need for the output json to be common to the two datatypes in the ndt-server. Especially if this is holding back development velocity for the NDT7 protocol, and does not break ndt5 backward compatibility (it does not).

yachang:
FYI, the removed field is result.Upload.ServerMeasurements.Elapsed, which block the ETL schema update code.

stephen-soltesz:
Done!

PR #271, opened 2020-03-16
Add separate structs for ndt5 & ndt7 datatypes

stephen-soltesz:
This change separates the NDTResult struct into two NDT5 and NDT7 types.

This change does not remove any fields from the new NDT7 struct, in order to remain maximally backward compatible with existing ndt7 archived data.

From adhoc testing in sandbox by @gfr10598, we expect that the etl parse (once updated to use this new ndt5result struct) should be able to continue writing to the existing tables. However, some manual steps will be necessary to fix the existing table schemas before `etl/cmd/update-schema` works as intended for the ndt5 tables.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/271)

### DIFF
--- a/data/result.go
+++ b/data/result.go
@@ -10,7 +10,7 @@ import (
 	"github.com/m-lab/ndt-server/ndt7/model"
 )
 
-// NDTResult is the struct that is serialized as JSON to disk as the archival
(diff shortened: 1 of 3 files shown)


stephen-soltesz:
Part of https://github.com/m-lab/ndt-server/issues/260

PR #273, opened 2020-03-20
Use inetdiag.BBRInfo as base of model.BBRInfo

stephen-soltesz:
This change merges the small files in ndt7/model into a single archivaldata.go file. And, this change modifies the custom BBRInfo struct to embed the inetdiag.BBRInfo struct to prevent ambiguous naming for the same information.

As a result of changing the `model.BBRInfo` struct, the `MaxBandwidth` field is removed in favor of the equivalent `BW` field from `inetdiag.BBRInfo`. The `inetdiag.BBRInfo` struct field names are faithfully preserve the underlying Linux struct field names.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/273)


stephen-soltesz:
FYI: @bassosimone -- if this has client-visible side-effects, please help me identify them. This makes the resulting schema easier to work with and similar to the tcpinfo schema in a helpful way.

bassosimone:
@stephen-soltesz did you check whether `spec/ndt7-protocol.go` needs to be changed/bumped?

stephen-soltesz:
@bassosimone I checked to see whether the ndt7 spec referenced the BBRInfo struct, and while it explicitly mentions several of the `Measurement` fields (AppInfo, TCPInfo, ConnectionInfo) it makes no reference to the BBRInfo struct / object. Was that intentional?

bassosimone:
@stephen-soltesz thanks for checking! I thought it was mentioned. I believe it was not mentioned because we didn't want to overpromise in the spec. In particular, it seems that we wanted to defend us against someone misunderstanding BW and using it without understanding its implications. (I mean, it is the kernel's estimate and may fluctuate a lot.) On this note, I quite like the new naming, i.e. just "BW", because it is probably less easy to be misunderstood than "MaxBandwidth".

PR #289, opened 2020-05-07
Add metrics shared with clients to ArchivalData

stephen-soltesz:
While working on the ndt5 parser migration to standard columns, I discovered that we are not collecting some metrics in the ndt5 archival structure.

This change adds two new values to the s2c and c2s archival data: LossRate and MinRTT respectively.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/289)


stephen-soltesz:
@pboothe the build may break due to upstream dependencies atm, but are there more things we should collect?

pboothe:
People want: bandwidth, latency, loss rate, jitter.  This is in order from more to less desireable, and from more to less well-defined.  If we report MinRTT and MaxRTT, then the people who want "jitter" can derive on reasonable version of that metric by calculating `MaxRTT - MinRTT`.

PR #290, opened 2020-05-12
reference client fixes

aweits:
two issues on the upload side:
1.) fix for webkit browsers in the upload loop
2.) grab the messages from the server during upload
also changed the timing to use performance.now()

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/290)


pboothe:
Got it. There are several problems here.

The biggest one is that the message size is held constant as the loop spins.  I can totally see how, with a small message size on a fast network, the loop will spin and never start buffering.  If you bump up the message size to start at, say, 1MB or even 10MB, does the webkit websocket buffer always stay at zero? I would expect it to fill up pretty quickly.

In that case, the secret is to change up the loop logic to get an appropriately-sized message every trip through the loop. Most of the time, that message will be the same as the previous one (it should be the same object, and cached so that the common case is just to return a cached reference), but every now and then it will increase in size and return a new object with an updated cache.  Exponential growth should quickly lead to full buffers.

Something like 
```js
const MessageProvider = function (startingSize, maxSize) {
  let cachedMessage = ...; // Declare a new messageof startingSize here
  const getMessage = function (bytesQueued) {
    if (bytesQueued > 16 * cachedMessage.length && cachedMessage.length < maxSize) {
       cachedMessage = ...; // Make a new message of double the size
    }
    return cachedMessage;
  };
  return {
    getMessage: getMessage
  }
}
```

which would then later be used in a loop like:

```js
const mp = MessageProvider(8192, 16777216);
let message = mp.getMessage(0);
let queuedBytes = 0;
while (sock.bufferedAmount < 4 * message.length) {
  message = mb.getMessage(queuedBytes);
  sock.send(message);
  queuedBytes += message.length;
}
```

aweits:
> Got it. There are several problems here.
> 
> The biggest one is that the message size is held constant as the loop spins. I can totally see how, with a small message size on a fast network, the loop will spin and never start buffering. If you bump up the message size to start at, say, 1MB or even 10MB, does the webkit websocket buffer always stay at zero? I would expect it to fill up pretty quickly.

In short, I concur on the several problems assessment. I'm just not sure there's an elegant solution that doesn't involve fixing the misreporting of buffering. My primary goal was to have this: http://smoke.rc.rit.edu:9080/static/ndt7_mine.html working for webkit on mobile devices.

If I set initialMessageSize to 1MB, and change the loop invariant to be "while ((sock.bufferedAmount == 0)) {", I never get an AppInfo message clientside, and almost immediately get a "{"fields":{"error":"websocket: close 1006 (abnormal closure): unexpected EOF"},"level":"warn","timestamp":"2020-06-11T12:58:13.344082382-04:00","message":"receiver: conn.ReadMessage failed"}" serverside.

If, on the other hand, I set initialMessageSize to 81 (!), I never get an AppInfo message and the websocket connection stays alive until the test completes. **I suspect that the websocket buffer is just not reported correctly (i.e. always 0), there is a standing queue in the buffer, and at some point overloading the buffer causes the connection to abort.** So my goal here was at least to exit the loop so that we honor the reporting interval.

As justification for my belief that there is indeed buffering taking place: with this patch, as-is, I see the following serverside during an upload test from my iPhone:

{"fields":{},"level":"debug","timestamp":"2020-06-11T13:10:26.488497047-04:00","message":"receiver: ApplicationLevel RTT: 324 ms"}
{"fields":{},"level":"debug","timestamp":"2020-06-11T13:10:26.674607406-04:00","message":"receiver: ApplicationLevel RTT: 485 ms"}
{"fields":{},"level":"debug","timestamp":"2020-06-11T13:10:26.95417329-04:00","message":"receiver: ApplicationLevel RTT: 714 ms"}
{"fields":{},"level":"debug","timestamp":"2020-06-11T13:10:27.592940503-04:00","message":"receiver: ApplicationLevel RTT: 1270 ms"}
{"fields":{},"level":"debug","timestamp":"2020-06-11T13:10:27.826103254-04:00","message":"receiver: ApplicationLevel RTT: 1420 ms"}
{"fields":{},"level":"debug","timestamp":"2020-06-11T13:10:27.826229382-04:00","message":"receiver: ApplicationLevel RTT: 1177 ms"}
{"fields":{},"level":"debug","timestamp":"2020-06-11T13:10:30.402977575-04:00","message":"receiver: ApplicationLevel RTT: 3378 ms"}
{"fields":{},"level":"debug","timestamp":"2020-06-11T13:10:32.718119994-04:00","message":"receiver: ApplicationLevel RTT: 5643 ms"}

pboothe:
Dangit! Frigging webkit.

Another solution proposed was to change the `while()` to an `if()` like in https://github.com/bassosimone/bassosimone.github.io/blob/master/ndt7/ndt7-core.js#L61

I really want the message size to be able to grow as quickly as possible, and your solution only allows it to grow once per reporting interval, which means that it can only double 4 times per second, and on fast networks it will need to double much faster than that. (on a 1 GB connection, messages should be 1-2 megabytes in size in order to fill the link, so we need to start doubling that 8k message quickly in order to grow that fast). 

Let's either take the solution from @bassosimone and bring it in here, or let's make the loop run until *either* the buffer is full *or* the message size should grow.  But I think the one that simplifies things to eliminate the inner loop is the right approach.  If you straight-up change the `while` to an `if` while keeping everything else the same as it was...

```js
    const underbuffered = 7 * data.length
    if (sock.bufferedAmount < underbuffered) {
      sock.send(data)
      total += data.length
    }
```

do things work well?

If not, then I think we need to add a third exit condition for the while loop, like:

```js
while ((sock.bufferedAmount < underbuffered) &&
	   (performance.now() - previous < every) &&
           (total < 16 * data.length)) {
```

The first is obviously simpler, but the second maybe is better because it precisely expresses our intent: just keep trying to fill the pipe until you have to do something else, and that something else could be: wait for the pipe to drain, send a message, or grow the message size.

pboothe:
The more I think about this, the happier I am with the three-condition while loop.  It is a direct expression of our intent. Is the three-condition loop okay with you?

aweits:
> The more I think about this, the happier I am with the three-condition while loop. It is a direct expression of our intent. Is the three-condition loop okay with you?

I find that needs to be stated like this to function properly, but this seems to work well:
 
    while ((sock.bufferedAmount < underbuffered) &&
	   (performance.now() - previous < every)) {
      sock.send(data)
      total += data.length
      if (data.length < maxMessageSize && data.length < (total - sock.bufferedAmount)/16) {
        break
      }
    }

PR #295, opened 2020-06-05
Eliminate saver logic and remove sender & receiver channels

stephen-soltesz:
This change is part of a sequence of changes that intend to make the ndt7 server logic transparently correct, easier to instrument with prometheus metrics, and unit testable.

This change eliminates the saver logic and many of the channels that obscured control flow.

In both cases, the local "sender" implementation allocates a "measuerer". The measurer controls the maximum lifetime of the measurement, as before. However, the timeouts are passed explicitly to `measurer.Start`, placing all related timeouts in the same calling level in `sender.Start`.

This change adds `measurer.Stop` to contain the logic to cleanly shutdown the measurer, rather than requiring calling logic to drain the channel.

Collection of the test `StartTime` and `EndTime` is now as close to the sender loop as possible.

Client and Server measurements are saved directly to the archival data parameter. In both cases, there is a single writer, and because `upload.Do` and `download.Do` remain synchronous operations, access by `writeResult` remains safe.

Future changes will eliminate the 'sender' packages within the upload and download packages in favor of combining them, so all logic is "in the same place".

This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/295)


stephen-soltesz:
https://github.com/m-lab/dev-tracker/issues/573

PR #299, opened 2020-06-15
Fix netx.Conn metrics and reduce ndt7 websocket buffer

stephen-soltesz:
During load testing, I discovered that ndt5 can close netx.Conns multiple times. This change guarantees that the metric is only decremented once.

During load testing, I discovered that ndt7 RAM requirements limit maximum concurrent connections to 1/3 ndt5's level with ndt-server RSS maxing out around 9.5GB before the system failed. This appears to be due to the extremely large websocket buffers. This change adds a new constant for the websocket buffer sizes of 1MB. Buffer sizes are independent of websocket message sizes.

With these settings load testing was able to comfortably support 300 concurrent connections (the previous water mark from NDT5) or ~1700 tests/min (very close to the theoretical limit of 1800 / min for 10sec tests, which is much tighter run time than ndt5's 15sec/test).

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/299)

### DIFF
--- a/ndt7/spec/spec.go
+++ b/ndt7/spec/spec.go
@@ -22,6 +22,13 @@ const MaxMessageSize = 1 << 24
 // a good compromise between Go and JavaScript as seen in cloud based tests.
 const MaxScaledMessageSize = 1 << 20
 
+// DefaultWebsocketBufferSize is the read and write buffer sizes used when
+// creating a websocket connection. This size is independent of the websocket
+// message sizes defined above (which may be larger) and used to optimize read
+// and write operations. However, larger buffers will practically limit the
+// total number of concurrent connections possible.
+const DefaultWebsocketBufferSize = 1 << 20
+
 // ScalingFraction sets the threshold for scaling binary messages. When
 // the current binary message size is <= than 1/scalingFactor of the
 // amount of bytes sent so far, we scale the message. This is documented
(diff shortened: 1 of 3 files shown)


stephen-soltesz:
https://github.com/m-lab/dev-tracker/issues/603

stephen-soltesz:
https://github.com/m-lab/dev-tracker/issues/573

netx/net.go
```diff
@@ -110,7 +112,7 @@ func (ln *Listener) Accept() (net.Conn, error) {
 // returned by LocalAddr and RemoteAddr should be released before calling Close.
 func (mc *Conn) Close() error {
 	mc.fp.Close()
-	CurrentOpenConns.Dec()
+	mc.once.Do(CurrentOpenConns.Dec)
```
pboothe, line 115: Thanks for this.

ndt7/spec/spec.go
```diff
@@ -22,6 +22,13 @@ const MaxMessageSize = 1 << 24
 // a good compromise between Go and JavaScript as seen in cloud based tests.
 const MaxScaledMessageSize = 1 << 20
 
+// DefaultWebsocketBufferSize is the read and write buffer sizes used when
+// creating a websocket connection. This size is independent of the websocket
+// message sizes defined above (which may be larger) and used to optimize read
+// and write operations. However, larger buffers will practically limit the
+// total number of concurrent connections possible.
+const DefaultWebsocketBufferSize = 1 << 20
```
pboothe, line 30: Would prefer a comment after this `// 1 MB` or maybe `// 1 MiB` if you are feeling pedantic.

pboothe:
It very much captures the spirit of my feedback. The other constants also need to have units, and this seemed as good a place to start as any. Now all the other constants (that aren't unitless or durations) look a little underdocumented, which is as it should be.  The next person who touches their comments will now (hopefully) put a little extra in there.

PR #303, opened 2020-06-25
update default webpage for ndt-server

critzo:
This PR updates the default server webpage to direct people who may find it to our main website. @pboothe  and @laiyi-ohlsen please take a look?

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/303)

### DIFF
--- a/ndt-server.go
+++ b/ndt-server.go
@@ -75,7 +75,7 @@ func defaultHandler(w http.ResponseWriter, req *http.Request) {
 
 <p>More information about <a href="https://measurementlab.net/tests/ndt/">NDT</a>, other M-Lab hosted <a href="https://measurementlab.net/tests/">tests</a>, the M-Lab platform, and the open data we publish can be found on our <a href="https://measurementlab.net">website</a>.</p>
 
-<p>If you need assistance, please email support@measurementlab.net, or visit <a href="https://measurementlab.net">our website</a> for complete information about the M-Lab platform, tests, and data.</p>
+<p>If you need assistance, please email support&#64;measurementlab.net, or visit <a href="https://measurementlab.net">our website</a> for complete information about the M-Lab platform, tests, and data.</p>
 </body>
 </html>
 `)))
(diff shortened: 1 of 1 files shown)


laiyi-ohlsen:
Do we want to add the support email as well?

pboothe:
I thought about that, and the helper links are incom

critzo:
Agreed @laiyi-ohlsen I will add the support email address. 

@stephen-soltesz It was my understanding that the links to the locally provided ndt5 and ndt7 pages were to be removed in favor of directing users to our website. The ndt7 test currently provided is the original prototype and should be replaced with @pboothe's `ndt7-js`. 

I'm not sure about the other urls. We might assume that only developers or people deploying ndt-server come to this page, but I don't think we have data about that. We could add a section that provides the urls in text format, and a link to the Developer resources page?  Something like:

"If you are deploying ndt-server in your own infrastructure, the server provides these endpoints for ndt5, ndt7, and ..."

critzo:
@pboothe or, if we put the URLs in a README for third parties who choose to deploy `ndt-server` with instructions on updating the default page if they choose to do so. Then we can have one default file and people deploying it themselves know where to send their folks.

critzo:
@laiyi-ohlsen my latest commit adds our support@ email address.

ndt-server.go
```diff
@@ -75,7 +75,7 @@ func defaultHandler(w http.ResponseWriter, req *http.Request) {
 
 <p>More information about <a href="https://measurementlab.net/tests/ndt/">NDT</a>, other M-Lab hosted <a href="https://measurementlab.net/tests/">tests</a>, the M-Lab platform, and the open data we publish can be found on our <a href="https://measurementlab.net">website</a>.</p>
 
-<p>Please visit https://measurementlab.net</a> for complete information about the M-Lab platform, tests, and data.</p>
+<p>If you need assistance, please email support@measurementlab.net, or visit <a href="https://measurementlab.net">our website</a> for complete information about the M-Lab platform, tests, and data.</p>
```
pboothe, line 78: Please change the @ sign to:
```html
	&#64;
```
which is rendered as @ but is less likely to be picked up by simple robots.  If someone has built a sophisticated robot, then they already know about webpages with that email address on them, so it's fine.

critzo: Addressed in latest commit.

critzo:
@laiyi-ohlsen Any final changes to recommend?

pboothe:
Lai Yi  said she is happy with this content. I would like to submit #305 instead of this one, which contains that same content, but put in the repo differently.

PR #333, opened 2021-06-09
Support cutomising TLS configuration

nosnilmot:
Add -tls_config command line option to allow specifying either "intermediate" or "modern" TLS configuration based on recommendations at https://wiki.mozilla.org/Security/Server_Side_TLS, allowing for disabling of older TLS versions.

CurveP384 is not included in CurvePreferences as there is no optimized implementation in Go.

Default configuration is unchanged, so if no -tls_config parameter is specified existing behavior is maintained.

I recognize this is somewhat unnecessary, given that there's no real risk from using TLS 1.0 for speed tests, but this helps tick boxes in corporate security compliance checklists.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/333)

### DIFF
--- a/ndt-server.go
+++ b/ndt-server.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"crypto/tls"
 	"flag"
 	"fmt"
 	"log"
@@ -37,6 +38,7 @@ var (
 	ndt5WssAddr       = flag.String("ndt5_wss_addr", ":3010", "The address and port to use for the ndt5 WSS test")
 	certFile          = flag.String("cert", "", "The file with server certificates in PEM format.")
 	keyFile           = flag.String("key", "", "The file with server key in PEM format.")
+	tlsConfig         = flag.String("tls_config", "old", "TLS configuration: modern, intermediate or old.")
 	dataDir           = flag.String("datadir", "/var/spool/ndt", "The directory in which to write data files")
 	htmlDir           = flag.String("htmldir", "html", "The directory from which to serve static web content.")
 	tokenVerifyKey    = flagx.FileBytesArray{}
@@ -96,9 +98,41 @@ func init() {
 
 // httpServer creates a new *http.Server with explicit Read and Write timeouts.
 func httpServer(addr string, handler http.Handler) *http.Server {
+	tlsconf := &tls.Config{}
+	// "modern" and "intermediate" are defined here:
+	// https://wiki.mozilla.org/Security/Server_Side_TLS
+	switch *tlsConfig {
+		case "modern":
+			tlsconf = &tls.Config{
+				MinVersion: tls.VersionTLS13,
+				CurvePreferences: []tls.CurveID{
+					tls.CurveP256,
+					tls.X25519,
+				},
+			}
+		case "intermediate":
+			tlsconf = &tls.Config{
+				MinVersion: tls.VersionTLS12,
+				CurvePreferences: []tls.CurveID{
+					tls.CurveP256,
+					tls.X25519,
+				},
+				CipherSuites: []uint16{
+					tls.TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,
+					tls.TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,
+					tls.TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305,
+					tls.TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305,
+					tls.TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,
+					tls.TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,
+					tls.TLS_RSA_WITH_AES_256_GCM_SHA384,
+					tls.TLS_RSA_WITH_AES_128_GCM_SHA256,
+				},
+			}
+		}
 	return &http.Server{
 		Addr:    addr,
 		Handler: handler,
+		TLSConfig: tlsconf,
 		// NOTE: set absolute read and write timeouts for server connections.
 		// This prevents clients, or middleboxes, from opening a connection and
 		// holding it open indefinitely. This applies equally to TLS and non-TLS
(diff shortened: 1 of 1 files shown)


nosnilmot:
The build "failure" is due to exceeding docker hub rate limits when creating the image:

`anyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit`

robertodauria:
Hi @nosnilmot, thank you for your contribution!

Do I understand correctly that this change is mostly for users who are deploying their own instance of ndt-server?

I agree with allowing users to tick boxes on corporate security compliance checklists, but I would propose implementing it as a simple `-tls1.3` flag to enforce TLS 1.3 where needed and leaving the choice to the `crypto/tls` library otherwise. 

When using TLS 1.3:

- the list of CipherSuites isn't configurable. The available options are:
```
	// TLS 1.3 cipher suites.
	TLS_AES_128_GCM_SHA256       uint16 = 0x1301
	TLS_AES_256_GCM_SHA384       uint16 = 0x1302
	TLS_CHACHA20_POLY1305_SHA256 uint16 = 0x1303
```
```
    // CipherSuites is a list of supported cipher suites for TLS versions up to
    // TLS 1.2. If CipherSuites is nil, a default list of secure cipher suites
    // is used, with a preference order based on hardware performance. The
    // default cipher suites might change over Go versions. Note that TLS 1.3
    // ciphersuites are not configurable.
```

- the preference order is [based on hardware performance](https://golang.org/pkg/crypto/tls/#Config). For instance, since Go1.16 the server [won't use AES](https://golang.org/doc/go1.16#crypto/tls) if the client does not signal a preference for it or the server itself doesn't have hardware support, which is the case for, e.g., ARMv7 CPUs without cryptographic extensions. By specifying a custom list, these automatisms are lost.

- `CurveP384` is enabled, but it also comes after both `X25519` and `CurveP256` [in the preference list](https://golang.org/src/crypto/tls/common.go) (`defaultCurvePreference` on line 978) meaning - if I understand correctly - it will only be used as a last resort.

As a general rule, we would like to take advantage of Go's `crypto/tls` library when it comes to determining which suite to use, as they are in a better position to assess which algorithms have better support/performance to be used preferentially at a given point in the library's development.

It's also true that, as you mentioned, the NDT5 and ndt7 protocols just send a stream of random bytes so there isn't anything confidential. The main reasons for using TLS are simulating a real-world condition (where most of the traffic is encrypted, nowadays) and preventing middleboxes from compressing the data in case a client doesn't randomize the message properly during the upload measurement.

Would a `-tls1.3` command-line flag as proposed above fit your use case? Or perhaps `-tls.version` which could be `1.2` or `1.3` and set the `MinVersion` field of `tls.Config` accordingly?

nosnilmot:
Hi @robertodauria, thanks for the feedback.

> Do I understand correctly that this change is mostly for users who are deploying their own instance of ndt-server?

Yes, that's correct, sorry for not being explicit about this originally. I assume public servers want to have the broadest possible client support, at least in the short term.

> When using TLS 1.3:

Snip lots of useful info. Good to see the cipher suite selection pretty much matches what I wanted anyway.

> As a general rule, we would like to take advantage of Go's `crypto/tls` library when it comes to determining which suite to use, as they are in a better position to assess which algorithms have better support/performance to be used preferentially at a given point in the library's development.

On cipher suite selection I agree, after this education.

> It's also true that, as you mentioned, the NDT5 and ndt7 protocols just send a stream of random bytes so there isn't anything confidential. The main reasons for using TLS are simulating a real-world condition (where most of the traffic is encrypted, nowadays) and preventing middleboxes from compressing the data in case a client doesn't randomize the message properly during the upload measurement.

Understood.

> Would a `-tls1.3` command-line flag as proposed above fit your use case? Or perhaps `-tls.version` which could be `1.2` or `1.3` and set the `MinVersion` field of `tls.Config` accordingly?

I'm not sure we can jump straight to 1.3 across the board yet, so would prefer the flexibility to specify minimum of either 1.2 or 1.3.

I think with these changes this PR might fit your suggestions?

- change flag from "-tls_config" to "-tls.version"
- change values from "intermediate, modern" to "1.2, 1.3"
- remove specification of preferred curves and cipher suites

If you agree, can you let me know if you prefer those changes as an additional commit to this branch/PR, or if you prefer them to be rebased and force-pushed (or in a new PR)?

robertodauria:
> * change flag from "-tls_config" to "-tls.version"
> * change values from "intermediate, modern" to "1.2, 1.3"
> * remove specification of preferred curves and cipher suites
> 
> If you agree, can you let me know if you prefer those changes as an additional commit to this branch/PR, or if you prefer them to be rebased and force-pushed (or in a new PR)?

Either adding new commits on top of this PR or creating a new one entirely (+ referencing this one for visibility into the discussion) sounds good to me. I like the `-tls.version=(1.2|1.3)` flag. Thanks! :)

nosnilmot:
I pushed the agreed changes in https://github.com/m-lab/ndt-server/pull/333/commits/db61651fd896dfb60c3880f2257d4a6342027392 is there anything else you want from me to be able to merge this?

ndt-server.go
```diff
@@ -96,9 +98,23 @@ func init() {
 
 // httpServer creates a new *http.Server with explicit Read and Write timeouts.
 func httpServer(addr string, handler http.Handler) *http.Server {
+	tlsconf := &tls.Config{}
+	// "modern" and "intermediate" are defined here:
+	// https://wiki.mozilla.org/Security/Server_Side_TLS
```
robertodauria, line 103: This comment is referring to the previous version of the code, could you please remove it?

robertodauria (changes requested):
Just a tiny change - other than that, LGTM.

ndt-server.go
nosnilmot: Oops, yes, removed in https://github.com/m-lab/ndt-server/pull/333/commits/3ede826aff4221c65ef1b91d9d15e614201b31b7

robertodauria (approved):
Thanks!

PR #343, opened 2021-12-03
Use a Reader in receiver.go to prevent unnecessary memory alloc

robertodauria:
This change is very similar to https://github.com/m-lab/ndt7-client-go/pull/63 - instead of using `conn.ReadMessage()` which allocates memory for the message, we get an `io.Reader`, check the message type and don't process it further if the type is not `websocket.TextMessage`.

Even if this is usually not a problem on M-Lab production hardware, this patch allowed me to measure a 1 Gb/s link fully with a ODROID XU-4 (armv7, 32 bit) running ndt-server, reducing the CPU usage from ~250% to ~120%

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/343)

### DIFF
--- a/ndt7/receiver/receiver.go
+++ b/ndt7/receiver/receiver.go
@@ -57,7 +57,7 @@ func start(
 		mtype, r, err := conn.NextReader()
 		if err != nil {
 			ndt7metrics.ClientReceiverErrors.WithLabelValues(
-				proto, string(kind), "read-message").Inc()
+				proto, string(kind), "read-message-type").Inc()
 			return
 		}
 		if mtype != websocket.TextMessage {
--- a/ndt7/receiver/receiver.go
+++ b/ndt7/receiver/receiver.go
@@ -5,6 +5,7 @@ package receiver
 import (
 	"context"
 	"encoding/json"
+	"io/ioutil"
 	"time"
 
 	"github.com/gorilla/websocket"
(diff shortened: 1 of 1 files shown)


stephen-soltesz (approved):
:lgtm: w/ one comment.

> *[Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/343#-:-Mq1S4fd42C2OlBhz4Re:b59tnm)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @robertodauria)

___
*[ndt7/receiver/receiver.go, line 60 at r1](https://reviewable.io/reviews/m-lab/ndt-server/343#-Mq1RrhcEGjNfb-266rR:-Mq1RrhcEGjNfb-266rS:br1wium) ([raw file](https://github.com/m-lab/ndt-server/blob/1b67be21242d424579947b75fad4924942f2bf79/ndt7/receiver/receiver.go#L60)):*
> ```Go
> 		if err != nil {
> 			ndt7metrics.ClientReceiverErrors.WithLabelValues(
> 				proto, string(kind), "read-message").Inc()
> ```

Please update this to `read-message-type`.

robertodauria:
Merging with a failing test (unrelated to this change):

```--- FAIL: Test_ContextCancelsMain (1.20s)```

PR #360, opened 2022-07-20
RFC: Add user-specified parameters for transfer byte limits

stephen-soltesz:
This change is a prototype that would add user parameters to NDT7 measurements and archival results. The current implementation is not ready to merge in its current form. This is an RFC.

The current implementation supports the following parameters:

* `param_close_after_download_bytes_acked` - the server will close the websocket conn after observing tcpinfo.BytesAcked greater than the given value.
* `param_close_after_upload_bytes_received` - the server will close the websocket conn after observing tcpinfo.BytesReceived greater than the given value.
* `param_cubic_download` - parsed but not enableable. Included as an illustration of how we could enable this or other options in the future.

In the first two cases, the measurement/observation events depend on the `measurer` memoryless [sample frequency][memoryless] - meaning that the time between actually reaching the `BytesAcked` or `BytesReceived` threshold (kernel state) and observing that (ndt-server) could be up to [MaxPoissonSamplingInterval](https://github.com/m-lab/ndt-server/blob/master/ndt7/spec/spec.go#L50) (625ms). So, there is a fundamental imprecision around when the server observes and closes the connection; there is no guarantee about how much data is actually transferred, only that it is *at least* the amount requested.

Currently, the parameter value is in bytes and unconstrained. Ideally, we would limit these values to a fixed set of thresholds, e.g. 1, 2.5, 5, 10, etc..

[memoryless]: https://github.com/m-lab/ndt-server/blob/master/ndt7/measurer/measurer.go#L79-L82

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/360)

### DIFF
--- a/html/ndt7-upload.js
+++ b/html/ndt7-upload.js
@@ -14,7 +14,8 @@ onmessage = function (ev) {
     let now = performance.now()
     const duration = 10000  // millisecond
     const every = 250  // millisecond
-    if (now - start > duration) {
+    // Run for duration or until server closes connection.
+    if (now - start > duration || sock.readyState == sock.CLOSED) {
       sock.close()
       return
     }
--- a/html/ndt7.html
+++ b/html/ndt7.html
@@ -24,8 +24,10 @@
 </head>
 <body>
   <div>
-    <div id='download' class='result row'>[Download]</div>
-    <div id='upload' class='result row'>[Upload]</div>
+    <div id='downloadShort' class='result row'>[Download (short)]</div>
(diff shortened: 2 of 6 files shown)


stephen-soltesz:
FYI: @bassosimone @robertodauria @mattmathis your comments/feedback are welcome. Regard everything as having a question at the end: "this is how I built a prototype; is there a better way to do this?"

stephen-soltesz:
Obsolete

Issue #361, opened 2022-07-25
ndt7: goroutine leak at the end of a measurement


robertodauria:
This was initially discovered by @phillipa by noticing that the difference between `EndTime` and `StartTime` in many ndt7 BigQuery rows is incompatible with the `MaxRuntime` (15s) parameter of ndt7 measurements. Upon further research, I have found some measurements that appear to have lasted 40+ days.

The measurer goroutine (in the `ServerMeasurements` field) ends after 10s and the `MeanClientMbps` in BigQuery is computed from those measurements -- so it appears to be correct -- but this is still an issue when doing data analysis (e.g. computing the measurement duration distribution).

For this to happen, the deferred function that sets `EndTime` at https://github.com/m-lab/ndt-server/blob/master/ndt7/upload/sender/sender.go#L47 must have been called late.

I've dumped the goroutines on a running production server (mlab1-del01) to check what's actually happening, and indeed there are receiver goroutines the have been stuck in `NextReader()` for minutes:
```

goroutine 2443141 [syscall, 9 minutes]:
syscall.Syscall(0x0, 0x23, 0xc00de67f5c, 0xc5ba, 0x4e5265, 0x7f2febd28090, 0x72)
	/usr/local/go/src/syscall/asm_linux_amd64.s:20 +0x5
syscall.read(0x23, 0xc00de67f5c, 0xc5ba, 0xc5ba, 0x0, 0x0, 0x0)
	/usr/local/go/src/syscall/zsyscall_linux_amd64.go:686 +0x5a
syscall.Read(...)
	/usr/local/go/src/syscall/syscall_unix.go:187
internal/poll.ignoringEINTRIO(...)
	/usr/local/go/src/internal/poll/fd_unix.go:581
internal/poll.(*FD).Read(0xc007c99000, 0xc00de67f5c, 0xc5ba, 0xc5ba, 0x0, 0x0, 0x0)
	/usr/local/go/src/internal/poll/fd_unix.go:162 +0x145
net.(*netFD).Read(0xc007c99000, 0xc00de67f5c, 0xc5ba, 0xc5ba, 0x5a8, 0x0, 0x0)
	/usr/local/go/src/net/fd_posix.go:55 +0x4f
net.(*conn).Read(0xc00026a4b8, 0xc00de67f5c, 0xc5ba, 0xc5ba, 0x0, 0x0, 0x0)
	/usr/local/go/src/net/net.go:183 +0x91
[...]

PR #362, opened 2022-07-27
Force conn.Close() after MaxRuntime in handler.go

robertodauria:
Related to #361. I'll add some more information about the issue there.

This PR changes how `conn.Close()` is called at the end of a measurement. Even if the current code _looks_ correct, we have seen many instances where the receiver and sender goroutines are stuck reading from / writing to the socket for a very long time after the end of a measurement. The deferred Close() in handler.go is supposed to force-close the connection when the receiver returns, but since the receiver is stuck trying to read from the socket this deferred function actually never gets called. Instead of using a deferred function, we start a separate goroutine that waits on a context timeout (or manual cancellation, when everything goes well and the sender goroutine cancels the context).

Assuming this approach looks fine to you, I would deploy this change on a small subset of servers and verify that no connections longer than 15s are recorded on these. The volume of connections exhibiting this issue is currently high enough that a couple of days should suffice to confirm. Unfortunately, I was not able to replicate it locally as it does never seem to happen under "normal" network conditions.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/362)

### DIFF
--- a/ndt7/handler/handler.go
+++ b/ndt7/handler/handler.go
@@ -106,11 +106,11 @@ func (h Handler) runMeasurement(kind spec.SubtestKind, rw http.ResponseWriter, r
 	var rate float64
 	if kind == spec.SubtestDownload {
 		result.Download = data
-		err = download.Do(req.Context(), conn, data)
+		err = download.Do(ctx, conn, data)
 		rate = downRate(data.ServerMeasurements)
 	} else if kind == spec.SubtestUpload {
 		result.Upload = data
-		err = upload.Do(req.Context(), conn, data)
+		err = upload.Do(ctx, conn, data)
 		rate = upRate(data.ServerMeasurements)
 	}
 
--- a/ndt7/handler/handler.go
+++ b/ndt7/handler/handler.go
@@ -74,7 +74,8 @@ func (h Handler) runMeasurement(kind spec.SubtestKind, rw http.ResponseWriter, r
 	// that under particular network conditions the connection can remain open
 	// while the receiver goroutine is blocked on a read syscall, long after
 	// the client is gone. This is a workaround for that.
-	ctx, _ := context.WithTimeout(req.Context(), spec.MaxRuntime)
+	ctx, cancel := context.WithTimeout(req.Context(), spec.MaxRuntime)
+	defer cancel()
 	go func() {
 		<-ctx.Done()
 		warnonerror.Close(conn, "runMeasurement: ignoring conn.Close result")
--- a/ndt7/handler/handler.go
+++ b/ndt7/handler/handler.go
@@ -2,6 +2,7 @@
 package handler
 
 import (
+	"context"
 	"fmt"
 	"net"
 	"net/http"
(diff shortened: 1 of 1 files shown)


stephen-soltesz (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/362)* status: 0 of 1 approvals obtained (waiting on @robertodauria)

___
*[`ndt7/handler/handler.go` line 81 at r1](https://reviewable.io/reviews/m-lab/ndt-server/362#-N8-546f4UB7u1nUmKrI:-N8-546gDCo-Qas8ALs0:b5yatyc) ([raw file](https://github.com/m-lab/ndt-server/blob/22c4957a5dbdc3331716b819e79a2976665f7275/ndt7/handler/handler.go#L81)):*
> ```Go
> 		<-ctx.Done()
> 		warnonerror.Close(conn, "runMeasurement: ignoring conn.Close result")
> 	}()
> ```

Is there any harm to keeping the `defer`? I'm imagining there could be a benefit. If the test returns for some reason, no need to wait to close the connection until MaxRuntime, we could run it immediately.

In general, it's better to call the `cancel` function to release resources. So, the defer could additionally call `cancel`, to free this goroutine early as well.

robertodauria (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/362)* status: 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[`ndt7/handler/handler.go` line 81 at r1](https://reviewable.io/reviews/m-lab/ndt-server/362#-N8-546f4UB7u1nUmKrI:-N8-UKzlDbKW8F8AzrHP:b-bueffh) ([raw file](https://github.com/m-lab/ndt-server/blob/22c4957a5dbdc3331716b819e79a2976665f7275/ndt7/handler/handler.go#L81)):*
<details><summary><i>Previously, stephen-soltesz (Stephen Soltesz) wrote…</i></summary><blockquote>

Is there any harm to keeping the `defer`? I'm imagining there could be a benefit. If the test returns for some reason, no need to wait to close the connection until MaxRuntime, we could run it immediately.

In general, it's better to call the `cancel` function to release resources. So, the defer could additionally call `cancel`, to free this goroutine early as well.
</blockquote></details>

You're right -- there is a comment in download.go and upload.go saying that
```
After timeout, the sender closes the conn, which results in receiver completing.
```

But the closer (called by sender) does not actually close the connection, it just sends a WebSocket control message and relies on the other end closing the connection: https://github.com/m-lab/ndt-server/blob/master/ndt7/closer/closer.go#L16

What do you think of this change that just cancels the context in the defer, instead of restoring the original `defer warnonerror.Close`?

stephen-soltesz (approved):
> *[Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/362)* status: :shipit: complete! 1 of 1 approvals obtained

___
*[`ndt7/handler/handler.go` line 81 at r1](https://reviewable.io/reviews/m-lab/ndt-server/362#-N8-546f4UB7u1nUmKrI:-N8-WDNc2TTtDmqNGdSJ:bc1r3cx) ([raw file](https://github.com/m-lab/ndt-server/blob/22c4957a5dbdc3331716b819e79a2976665f7275/ndt7/handler/handler.go#L81)):*
<details><summary><i>Previously, robertodauria (Roberto D'Auria) wrote…</i></summary><blockquote>

You're right -- there is a comment in download.go and upload.go saying that
```
After timeout, the sender closes the conn, which results in receiver completing.
```

But the closer (called by sender) does not actually close the connection, it just sends a WebSocket control message and relies on the other end closing the connection: https://github.com/m-lab/ndt-server/blob/master/ndt7/closer/closer.go#L16

What do you think of this change that just cancels the context in the defer, instead of restoring the original `defer warnonerror.Close`?
</blockquote></details>

Even better :+1:

PR #365, opened 2022-08-02
Lowering max message size in upload worker

mkwiatkowskisoldevelo:
We've discovered, that while using the embedded js client to run ndt7 tests, the upload speeds are very low for the connections at higher speeds (we were testing it with 4G and 10G). The reason is that the WebSocket connection is closed at the begging of the tests. The connection is closed by the server because its buffer is overflowing. This happens on the following browsers: Chrome, Edge, and Safari. Firefox is completing the upload test normally for the same set-up.

As a fix for this issue, we're proposing lowering the maxMessageSize on the upload worker on the JS client from 16MB to 8MB. This is also a value used in the other JS client: [ndt7-js](https://github.com/m-lab/ndt7-js). This client could also replace the current one, if it's OK with you I could add it as a part of this PR. I see that there is a [PR](https://github.com/m-lab/ndt-server/pull/291) that introduces that client, but it is over 2 years old and there were some changes in the other repository recently.

Another way to stop the upload test from failing is to slightly increase MaxMessageSize on the server side, which is set to 16MB. Changing it to 18MB is enough for our testing.

It would help us greatly in using NDT if you could include this change or resolve this issue.

Thank you very much in advance.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/365)

### DIFF
--- a/html/ndt7-upload.js
+++ b/html/ndt7-upload.js
@@ -18,7 +18,7 @@ onmessage = function (ev) {
       sock.close()
       return
     }
-    const maxMessageSize = 16777216 /* = (1<<24) = 16MB */
+    const maxMessageSize = 8388608 /* = (1<<2) = 8MB */
     if (data.length < maxMessageSize && data.length < (total - sock.bufferedAmount)/16) {
       data = new Uint8Array(data.length * 2) // TODO(bassosimone): fill this message
     }
--- a/html/ndt7-upload.js
+++ b/html/ndt7-upload.js
@@ -18,7 +18,7 @@ onmessage = function (ev) {
       sock.close()
       return
     }
-    const maxMessageSize = 8388608 /* = (1<<2) = 8MB */
+    const maxMessageSize = 8388608 /* = (1<<23) = 8MB */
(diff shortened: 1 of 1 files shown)


stephen-soltesz:
@mkwiatkowskisoldevelo thank you for this change. @robertodauria is currently on vacation through Aug 15th and has the most experience with performance tuning the server and js clients. He is the best to review your proposed changes.

robertodauria (approved):
:lgtm:

As far as I know the minimal client included in this repository is only used for e2e testing via nodejs, while ndt7-js is the recommended reference client for integrators. In ndt7-js I changed the maximum message size to 8MB some time ago to work around a different bug (only happening on Safari). It makes sense to also change it here, especially if it's causing problems on other browsers, too.

> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/365#-:-N9gLDh_4dJXyelE49zs:b-myivh7)* status: :shipit: complete! 1 of 1 approvals obtained

PR #367, opened 2022-08-04
Added parameter for changing server side max messasge size for the NDT upload test

mkwiatkowskisoldevelo:
We've added this change to allow changing the max accepted message size for the upload test on the server side. This property might be influencing the recorded performance so we think that this might be a good addition.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/367)

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,8 @@ docker run --network=bridge                \
            -key /certs/key.pem             \
            -datadir /datadir               \
            -ndt7_addr :4443                \
-           -ndt7_addr_cleartext :8080
+           -ndt7_addr_cleartext :8080      \
+           -ndt7_max_msg_size 16777216
 ```
 
 ### Alternate setup & running (Windows & MacOS)
--- a/ndt-server.go
+++ b/ndt-server.go
@@ -35,6 +35,7 @@ var (
 	// Flags that can be passed in on the command line
 	ndt7Addr          = flag.String("ndt7_addr", ":443", "The address and port to use for the ndt7 test")
 	ndt7AddrCleartext = flag.String("ndt7_addr_cleartext", ":80", "The address and port to use for the ndt7 cleartext test")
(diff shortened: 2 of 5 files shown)


robertodauria (commented):
Hi, thanks for contributing! :slightly_smiling_face:

Do you happen to have any performance data related to sending messages larger than 16MB? The two clients I worked on (ndt7-client-go and ndt7-js) use 1MB and 8MB respectively. The Go client didn't seem to get any performance improvement by sending > 1MB messages, while 8MB were used on the JavaScript client mostly just to be kind to the JS engine (i.e. return control to the event loop more often in case the sender code is running on the main thread).

ndt-server accepts messages up to 16MB (current MaxMessageSize) but sends messages [up to 1MB](https://github.com/m-lab/ndt-server/blob/main/ndt7/spec/spec.go#L23), mirroring what the Go client does.

I'm not opposed in principle to maxing MaxMessageSize easier to change without having to recompile ndt-server if there's a use case for that. Could you tell me more about the use case?

P.S.: Please note that according to the [ndt7 specification](https://github.com/m-lab/ndt-server/blob/main/spec/ndt7-protocol.md#websocket-channel-usage) right now `1<<24` seems to be the only MaxMessageSize possible since a binary message cannot exceed it:

```
An implementation MAY choose to limit the maximum WebSocket message size, but such limit MUST NOT be smaller than 1 << 24 bytes.

[...]

Binary messages MUST contain between 1 << 10 and 1 << 24 bytes
```

We're open to updating the spec, if the use case justifies it.

> *[Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/367#-:-N9gNFyXBK573EOFgw2P:b39jnii)* status: 0 of 1 approvals obtained

robertodauria:
Closing due to inactivity.

PR #379, opened 2023-01-31
Use performance.now() where available

robertodauria:
This PR changes the NDT5 client code to use the monotonic performance.now() instead of Date.now() where available. This fixes a rare bug that happens when the system clock is changed while a measurement is running, which would cause an incorrect client-side throughput to be displayed -- in extreme cases, even a negative value.

(The code [is basically the same](https://github.com/m-lab/ndt7-js/blob/main/src/ndt7-upload-worker.js#L7) as in ndt7-js)

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/379)

### DIFF
--- a/html/ndt-browser-client.js
+++ b/html/ndt-browser-client.js
@@ -8,6 +8,14 @@
 
 'use strict';
 
+// Prefer performance.now() over Date.now() since the former is monotonic.
+var now;
+if (typeof performance !== 'undefined' && typeof performance.now === 'function') {
+  now = () => performance.now();
+} else {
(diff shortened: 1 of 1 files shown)


stephen-soltesz (approved):
:lgmt:

> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/379#-:-NN7dn-34DhiB1UpFdX_:bnfozmr)* status: :shipit: complete! 1 of 1 approvals obtained

PR #396, opened 2023-11-03
Add optional eventsocket Server to ndt-server

stephen-soltesz:
This change adds an `eventsocket.Server` to the ndt-server for the ndt7 protocol.

With this change, the ndt-server could operate as a replacement for tcp-info's eventsocket for deployments that run with standard sidecar services (e.g. uuid-annotator, or others) but without needing tcp-info to be the event source. In this configuration, the ndt-server itself becomes the event source, limiting events to verified ndt7 measurements (rather than every TCP connection).

For this change to be possible, the ndt7 `handler.Handler` methods now use a pointer receiver so that the `Handler.Events` field can be reset after `ndt7test.NewNDT7Server` creates a server using the default `eventsocket.NullServer`. As well, we add a `handler_integration_test.go` to allow import of the existing `ndt7test` package, which sets up an ndt7 server for testing.

At the same time, I've removed use of the deprecated `"io/ioutil"` `TempDir` in favor of `t.TempDir()`.

Default server behavior is unchanged. This change adds a new flag to the ndt-server, `-tcpinfo.eventsocket string`. We reuse the existing eventsocket flag name, just as the sidecars do, to preserve the symmetry between the serving flag and the client flag names. Since we now have multiple event servers, ideally the `eventsocket` package would be moved to a generic location such as `m-lab/go/eventsocket` and imported by tcp-info and ndt-server, using a more general flag name in all cases and sidecars, e.g. `-server.eventsocket`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/396)

### DIFF
--- a/ndt-server.go
+++ b/ndt-server.go
@@ -26,6 +26,7 @@ import (
 	"github.com/m-lab/ndt-server/ndt7/spec"
 	"github.com/m-lab/ndt-server/platformx"
 	"github.com/m-lab/ndt-server/version"
+	"github.com/m-lab/tcp-info/eventsocket"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
@@ -193,6 +194,14 @@ func main() {
 		rtx.Must(err, "Failed to load verifier for when tokens are required")
 	}
 
+	// Make and start the event server.
+	eventSrv := eventsocket.NullServer()
+	if *eventsocket.Filename != "" {
+		eventSrv = eventsocket.New(*eventsocket.Filename)
+	}
+	rtx.Must(eventSrv.Listen(), "Could not listen on", *eventsocket.Filename)
+	go eventSrv.Serve(ctx)
+
 	// Enforce tokens and tx controllers on the same ndt5 resource.
 	// NOTE: raw ndt5 requests cannot honor tokens or differentiate between upload/downloads.
 	ndt5Paths := controller.Paths{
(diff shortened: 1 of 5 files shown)


cristinaleonr (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/396#-:-NiaWUTx6kGNTNgEYU2G:bg7igkv)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[`ndt7/handler/handler.go` line 117 at r2](https://reviewable.io/reviews/m-lab/ndt-server/396#-NiaWKJy1oEmvb5_-5xm:-NiaWKJy1oEmvb5_-5xn:b4ojev2) ([raw file](https://github.com/m-lab/ndt-server/blob/8a33ef48335004db53c66b6400773a3c31d31127/ndt7/handler/handler.go#L117)):*
> ```Go
> 	h.Events.FlowCreated(result.StartTime, data.UUID, id)
> 
> 	// Guarantee results are written even if test functions panics.
> ```

panic~~s~~?

stephen-soltesz (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/396)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[`ndt7/handler/handler.go` line 117 at r2](https://reviewable.io/reviews/m-lab/ndt-server/396#-NiaWKJy1oEmvb5_-5xm:-NiaXEcWAbSuqbJLqIGj:byvnd9s) ([raw file](https://github.com/m-lab/ndt-server/blob/8a33ef48335004db53c66b6400773a3c31d31127/ndt7/handler/handler.go#L117)):*
<details><summary><i>Previously, cristinaleonr (Cristina Leon) wrote…</i></summary><blockquote>

panic~~s~~?
</blockquote></details>

Thank you. Fixed.

PR #397, opened 2023-11-07
Fix: deterministic check for FlowDeleted

stephen-soltesz:
The previous test implementation was flaky. This change uses a channel to signal that the `FlowDeleted` path has been reached.

While the `srv.Close()` will prevent new connections, and the `downloadHelper()` closes the client connection, there is still an independent handler goroutine that reaches the `FlowDeleted` path after a short delay. Rather than using a larger static delay, this change sets a maximum delay and waits for the `deleted` channel to close, allowing the test to stop as quickly as it's able to.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/397)

### DIFF
--- a/ndt7/handler/handler_integration_test.go
+++ b/ndt7/handler/handler_integration_test.go
@@ -18,7 +18,7 @@ import (
 // fakeServer implements the eventsocket.Server interface for testing the ndt7 handler.
 type fakeServer struct {
 	created int
-	deleted int
+	deleted chan bool
 }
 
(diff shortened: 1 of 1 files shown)


cristinaleonr (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/397#-:-NieT62PEHXxzZGFmgAQ:bg7igkv)* status: :shipit: complete! 1 of 1 approvals obtained

PR #400, opened 2024-03-18
avoid mixed content breaking when using https

KwadroNaut:
When using the default ndt7.html that's shipped here, you'll end up with non-functional wss (or https). This makes it protocol agnostic.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/400)


robertodauria (approved):
Thanks!

PR #403, opened 2024-05-17
feat(ndt7): Use X-Forwarded-For hint for the client's IP address

auvik-bheesham:
When running behind a proxy the RemoteAddr in various places won't return the real client's IP. This commit uses various hints to get it, falling back to using the default.

cc @pthankappan

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/403)

### DIFF
--- a/ndt7/handler/handler.go
+++ b/ndt7/handler/handler.go
@@ -9,6 +9,7 @@ import (
 	"net/url"
 	"regexp"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/gorilla/websocket"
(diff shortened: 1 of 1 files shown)


auvik-bheesham:
Example: 

```
; go run ./cmd/ndt7-client/main.go -service-url 'wss://speedtest.example.com/ndt/v7/upload'
upload in progress with speedtest.example.com
Avg. speed  :   114.6 Mbit/s
upload: complete

Test results

    Server: speedtest.example.com
    Client: 100.96.17.113

              Upload
     Throughput:   114.6 Mbit/s
        Latency:     0.6 ms
```

The `Client: 100.96.17.113` is incorrect -- we're running the server in Kubernetes behind various proxies, so it's not quite returning the expected value.

auvik-bheesham:
Retracted for now since I noticed a couple of other places where `RemoteAddr` was being used.

PR #404, opened 2024-05-21
feat(ndt7): Use X-Forwarded-For hint for the client's IP address

auvik-bheesham:
When running behind a proxy the RemoteAddr in various places won't return the real client's IP. This commit uses various hints to get it, falling back to using the default.

cc @pthankappan

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/404)

### DIFF
--- a/ndt7/handler/handler.go
+++ b/ndt7/handler/handler.go
@@ -9,6 +9,7 @@ import (
 	"net/url"
 	"regexp"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/gorilla/websocket"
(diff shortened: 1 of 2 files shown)


robertodauria:
Hello, thank you for submitting this PR.

Unfortunately, for ndt-server to function properly, it requires an unterminated end-to-end TCP connection. The use of a reverse proxy, which would terminate the connection, is not supported. This will lead to inaccuracies in the TCPInfo, BBRInfo, and other network-related data, as they would be associated with the connection between the server and the reverse proxy instead of the server and the client.

Also note that if you are running the fullstack ndt-server, which includes sidecar services, all of them will still record the proxy's IP rather than the client's.

Additionally, data transmission to or from the client would be handled by the proxy's network stack, rather than the NDT server's network stack, resulting in inaccurate application-level metrics.

For these reasons, we recommend avoiding the use of ndt-server behind a reverse proxy and not overwriting the ClientIP field.

auvik-bheesham:
Hey! Thanks for the detailed response, I let the team know. Lots of new info here for me.

We'll reconsider our approach. We might end up finding some middle ground here, unsure. We've already got a bunch of infra set up and some requirements along with that, so there's going to be some acrobatics required to fit this in.

Either way, we learned a lot. Thanks again!

PR #408, opened 2025-01-08
Make writeResult errors fatal

robertodauria:
Some time ago we had an incident where NDT results were not written on disk due to permission issues, and even if ndt-server was logging these warnings no one noticed for a while. More recently, a BYOS pilot node was failing to write files due to a full disk.

An ndt-server instance that cannot write to disk is not particularly useful, so I propose that we consider this error fatal. In our k8s cluster, this means ndt pods will restart, triggering alerts and making the issue evident. In BYOS nodes, we'll still rely on the host noticing the issue, but the docker compose will stop.

### DIFF
--- a/ndt7/handler/handler.go
+++ b/ndt7/handler/handler.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/m-lab/access/controller"
 	"github.com/m-lab/go/prometheusx"
+	"github.com/m-lab/go/rtx"
 	"github.com/m-lab/go/warnonerror"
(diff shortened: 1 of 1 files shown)


cristinaleonr (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/408#-:-OG5E73-AXJRv68S-gFv:bg7igkv)* status: <span>:shipit:</span> complete! 1 of 1 approvals obtained

PR #410, opened 2025-04-17
Don't double sizes for slow clients

pboothe:
As discussed in Slack:

@pboothe : So what we need is to not double if the most recent tcpinfo data suggests that doubling will cause the next message to be too big?

@robertodauria : basically, yes. I was imagining it could be based on the actual delay between the tcpinfo snapshot and when the corresponding write to the socket returns -- if a tcpinfo snapshot couldn't be delivered for >= MaxPoissonSamplingInterval stop doubling the message size, for example, but other solutions are definitely possible

...or: setting a cap every time a tcpinfo snapshot is taken based on the current rate, check against this cap in sender.go before doubling, and keep the cap updated with each subsequent tcpinfo snapshot

@pboothe : Depending on what thread the tcpinfo measurement happens in, that last one feels like the right solution.

Nice and simple. Every measurement provides a max rate. Don’t double if the measured max rate suggests that each message will take more than half a second. Once TCP is cruising, we should cruise as well.

...

@robertodauria : half a second, or MaxPoissonSamplingInterval (600ms currently IIRC)? 

👍

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/m-lab/ndt-server
 
-go 1.20
+go 1.23
 
 require (
 	github.com/apex/log v1.9.0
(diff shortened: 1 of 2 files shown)


robertodauria:
Closed due to inactivity - reopen if needed.

PR #424, opened 2026-02-20
Skip Linux-specific c2s tests on non-Linux platforms

sahillllllllll-bit:
Some tests in ndt5/c2s rely on Linux-specific TCP features (e.g., TCP_INFO) and assume Linux kernel behavior. When executed on non-Linux platforms such as Windows, these tests hang or time out instead of failing fast.

This PR adds a runtime OS guard to skip these tests when `runtime.GOOS != "linux"`.

 ## Problem

Running `go test ./...` on Windows resulted in:

- 10-minute test timeout in `Test_DrainForeverButMeasureFor_NormalOperation`
- Hanging behavior due to missing Linux TCP features
- Poor contributor experience on non-Linux systems
- The tests depend on Linux-specific socket metrics that are not available on Windows.

## Solution

Add:
```
if runtime.GOOS != "linux" {
    t.Skip("Skipping test: requires Linux TCP features")
}
```
to the affected tests in `ndt5/c2s/c2s_test.go`.

This ensures:

- Tests run fully in Linux (CI environment)
- Non-Linux contributors do not experience long hangs
- Clear indication that the test is platform-specific

Impact

- Improves cross-platform developer experience
- Prevents long timeouts during local testing
- Maintains full coverage in Linux CI
- Does not change runtime or production behavior

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/424)

### DIFF
--- a/ndt5/c2s/c2s_test.go
+++ b/ndt5/c2s/c2s_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"net"
 	"net/http"
+	 "runtime"
 	"strconv"
 	"testing"
 	"time"
@@ -33,6 +34,9 @@ func MustMakeNetConnection(ctx context.Context) (protocol.MeasuredConnection, ne
 }
 
 func Test_DrainForeverButMeasureFor_NormalOperation(t *testing.T) {
+	 if runtime.GOOS != "linux" {
+        t.Skip("Skipping test: requires Linux TCP features")
+    }
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 	sConn, cConn := MustMakeNetConnection(ctx)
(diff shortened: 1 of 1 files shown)


sahillllllllll-bit:
@robertodauria please review it if you have time :)

robertodauria:
ndt-server is a Linux-only server: ndt7 requires kernel TCP_INFO and BBR support, neither of which is available on Windows. NDT5 is deprecated and while still supported on legacy platform servers, we are not accepting further changes to it. There is no use case for running these tests on Windows, so this change addresses a problem that doesn't exist.

The code also has inconsistent indentation (extra leading spaces on the import and conditionals).

We've been receiving a large volume of unsolicited PRs with no prior discussion or contact, and as a small team this is not sustainable. We are updating our contribution requirements. Going forward, please coordinate with a maintainer before opening PRs or issues. See https://github.com/m-lab/gsoc#2-engage-with-us for details.

sahillllllllll-bit:
@robertodauria Thank you for the clarification.

I initially started working on ndt-server from a Windows environment and encountered the 10-minute test hang while running `go test ./.... ` Since I did not immediately realize the Linux-only requirement at the kernel feature level (TCP_INFO / BBR), I investigated the issue and later shifted to a Linux environment.

My intention with the change was simply to prevent other new contributors from spending time debugging similar hangs if they happen to start from Windows before realizing the platform requirement. I understand that the project is explicitly Linux-only and that NDT5 is deprecated, so I respect the decision not to accept the change if it falls outside the intended scope.

If you believe the change could still be useful, I would be happy to correct the indentation and formatting issues and address any other concerns. If not, I completely understand and appreciate the clarification.

Thank you again for your guidance.

PR #426, opened 2026-02-27
fix(ndt7-protocol.md): clarify test "runtime" concept

bassosimone:
Existing historical clients (e.g., the minimal JavaScript client and the minimal Go client) did not have a consistent definition of the runtime, with the JavaScript client erring on the side of including the TCP+TLS+WebSocket handshake time and the Go client, instead, erring on the side of not including it.

This diff clarifies the matter. After a conversation with @robertodauria, and taking into account https://github.com/m-lab/ndt7-js/pull/111 and https://github.com/m-lab/msak-js/pull/23, we determined that it is probably better to adopt the behavior originally implemented by the minimal Go client.

The rationale for doing this is fairness. The original reason for killing a test unconditionally after ten seconds in JavaScript was grounded in UX concerns (do not keep the user waiting for more than 20 seconds overall). However, with high RTTs, this also means providing a less useful overall goodput estimate. To remediate this, we allow the client to run for more time.

Refer to the aforementioned pull requests for a more-in-depth analysis of this behavior, which helps reconstructing the chain of thoughts that led us to implement this spec change.

Since this diff is a spec clarification that correctly defines the concept of runtime, I am opting to bump the minor version of the spec.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/426)

### DIFF
--- a/spec/ndt7-protocol.md
+++ b/spec/ndt7-protocol.md
@@ -7,7 +7,7 @@ protocol](https://github.com/ndt-project/ndt). Ndt7 is based on
 WebSocket and TLS, and takes advantage of TCP BBR, where this
 flavour of TCP is available.
 
-This is version v0.10.0 of the ndt7 specification.
+This is version v0.11.0 of the ndt7 specification.
 
 ## Design choices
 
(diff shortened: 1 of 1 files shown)


robertodauria (approved):

bassosimone:
Note that `coveralls` is down due to technical issues.

PR #430, opened 2026-04-20
feat(ndt7): write integration claims (int_id, key_id) to ClientMetadata

robertodauria:
Reads int_id/key_id from verified access tokens (via the generic custom-claim
API from m-lab/access v0.1.0) and appends them to NDT7 result ClientMetadata.

Query-string values for int_id and key_id are filtered out so that only the
verified token is authoritative.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/430)


ndt7/handler/handler.go
```diff
@@ -263,6 +264,24 @@ func appendClientMetadata(data *model.ArchivalData, values url.Values) {
 	}
 }
 
+// appendIntegrationMetadata extracts integration claims from the request context
+// and appends them to ClientMetadata. These claims (int_id, key_id) are set by
+// the access token controller when the token contains integration-specific fields.
+func appendIntegrationMetadata(data *model.ArchivalData, ctx context.Context) {
+	ic, _ := controller.GetCustomClaim(ctx).(*IntegrationClaims)
+	if ic == nil {
+		return
```
bassosimone, line 273: Do you think we should log this error? (IIUC, it happens when we cannot type cast.)

ndt7/handler/handler.go
```diff
@@ -263,6 +264,24 @@ func appendClientMetadata(data *model.ArchivalData, values url.Values) {
...
+	}
+	if ic.IntegrationID != "" {
+		data.ClientMetadata = append(data.ClientMetadata,
+			metadata.NameValue{Name: "int_id", Value: ic.IntegrationID})
+	}
+	if ic.KeyID != "" {
+		data.ClientMetadata = append(data.ClientMetadata,
+			metadata.NameValue{Name: "key_id", Value: ic.KeyID})
+	}
+}
```
bassosimone, line 283: Do we want to support the case where just one of the two entries is empty? Your reasoning here seems to be to be "open" the incoming request only containing one of them. Is this something we should target or do you feel like we should be strict instead? What was your reasoning here for allowing either of them?

ndt7/handler/integration_claims.go
```diff
@@ -0,0 +1,9 @@
+package handler
```
bassosimone, line 1: IIUC Go uses `_foo` as something that is compiled when the `foo` tag is present. Maybe it is more robust to avoid using the `_foo` pattern when naming files and call this `integrationclaims.go`?

bassosimone (commented):
I did a pre-review of this in-progress PR and provided some comments. Apart from the identified comments, I did not have any other issue to flag.

ndt7/handler/handler.go
robertodauria: "If it's there, save it with the test results". This allows a policy change upstream (in case we do not want to include the key_id anymore, e.g.) without requiring an ndt-server change at the same time.

ndt7/handler/handler.go
robertodauria: Good point — distinguished now. The common case (no custom claim attached) returns silently, while a mismatched type (programming/config error: factory installed but a different type in context) emits a warning. Pushed in e0f8a8a.

ndt7/handler/integrationclaims.go
```diff
@@ -0,0 +1,9 @@
+package handler
```
robertodauria: Renamed to integrationclaims.go in e0f8a8a to stay clear of the *_GOOS.go convention.

robertodauria:
@claude

ndt7/handler/handler.go
bassosimone: Worth documenting the reasoning explicitly.

go.sum
```diff
@@ -55,8 +55,8 @@ github.com/kr/text v0.2.0 h1:5Nx0Ya0ZqY2ygV366QzturHI13Jq95ApcVaJBhpS+AY=
 github.com/kr/text v0.2.0/go.mod h1:eLer722TekiGuMkidMxC/pM04lWEeraHUUmBw8l2grE=
 github.com/kylelemons/godebug v1.1.0 h1:RPNrshWIDI6G2gRW9EHilWtl7Z6Sb1BR0xunSBf0SNc=
 github.com/kylelemons/godebug v1.1.0/go.mod h1:9/0rRGxNHcop5bhtWyNeEfOS8JIWk580+fNqagV/RAw=
-github.com/m-lab/access v0.0.13 h1:ZCtOEZp2qmcp0CFUBP21r42miNqqWu7gQVa8GxzASts=
-github.com/m-lab/access v0.0.13/go.mod h1:V3hWcsesp1kDgIRQOUXRU8uDTfUhYv/D0SJ4g3WOUS8=
+github.com/m-lab/access v0.0.14-0.20260420125445-78d3830cab5f h1:2wS199sy1CTyr9z/YQM8MfKRu8NeQeiZcinHlLfkp/w=
+github.com/m-lab/access v0.0.14-0.20260420125445-78d3830cab5f/go.mod h1:V3hWcsesp1kDgIRQOUXRU8uDTfUhYv/D0SJ4g3WOUS8=
```
bassosimone: @robertodauria for merging, it would be cool if we could pin to a stable version 🙏

robertodauria:
- Pinned m-lab/access to v0.1.0 now that it's been released
- Prevent int_id/key_id from being injected by the client via querystring parameters

ndt7/handler/handler.go
```diff
@@ -245,7 +246,7 @@ func downRate(m []model.Measurement) float64 {
 }
 
 // excludeKeyRe is a regexp for excluding request parameters from client metadata.
-var excludeKeyRe = regexp.MustCompile("^server_")
+var excludeKeyRe = regexp.MustCompile("^server_|^int_id$|^key_id$")
```
bassosimone, line 252: ❤️

ndt7/handler/handler.go
bassosimone: Can you document the rationale of why you're doing this? Just one line?

ndt7/handler/handler_test.go
```diff
@@ -2,14 +2,83 @@
...
+			name:      "with-int-id-only",
+			claim:     &IntegrationClaims{IntegrationID: "test-int"},
+			wantLen:   1,
+			wantIntID: "test-int",
+		},
+		{
+			name:    "unexpected-claim-type",
+			claim:   &struct{ Foo string }{Foo: "bar"},
+			wantLen: 0,
+		},
```
bassosimone, line 57: Can you add a test case in which the claims also appear on the query string so we verify that they are filtered out as it ought to be, please?

ndt7/handler/handler.go
```diff
@@ -263,6 +264,29 @@ func appendClientMetadata(data *model.ArchivalData, values url.Values) {
 	}
 }
 
+// appendIntegrationMetadata extracts integration claims from the request context
+// and appends them to ClientMetadata. These claims (int_id, key_id) are set by
+// the access token controller when the token contains integration-specific fields.
+func appendIntegrationMetadata(data *model.ArchivalData, ctx context.Context) {
```
bassosimone: Please, idiomatically swap the arguments and make `ctx` the first argument.

ndt7/handler/handler.go
```diff
@@ -109,6 +109,7 @@ func (h *Handler) runMeasurement(kind spec.SubtestKind, rw http.ResponseWriter,
 
 	// Collect most client metadata from request parameters.
 	appendClientMetadata(data, req.URL.Query())
+	appendIntegrationMetadata(data, req.Context())
```
bassosimone: Please, remember to update the PR description (we're not pinning a temporary commit anymore)

ndt7/handler/handler.go
bassosimone: Worth updating the comment to explain _why_ we're also excluding `int_id` and `key_id`.

ndt7/handler/handler.go
bassosimone: Also, can you document the rationale of pragmatically _overloading_ ClientMetadata to include not only entries supplied via the query string but also via the token? Do we need to update the documentation of the public field to make this explicit? (I guess so?)

bassosimone (approved):
🐳 This is a conditional approval. This PR is approved if you address all the comments I made (as well as the points raised by Claude) and addressing them is trivial enough that you don't feel like you need another round of review.

ndt7/handler/handler_test.go
bassosimone: Ah also a test case where only the key-id appears? (for symmetry) :pray:

robertodauria:
@claude check if the comments above have all been addressed + review again

PR #436, opened 2026-07-28
fix(bbr): remove cgo dependency, build statically with CGO_ENABLED=0

robertodauria:
Replace the cgo-based TCP_CC_INFO getsockopt in bbr_linux.go with a pure-Go implementation using golang.org/x/sys/unix, matching the pattern already used by tcpinfox. Build with CGO_ENABLED=0 so binaries are fully static, and drop gcc/linux-headers/musl-dev from the builder image along with the now-redundant netgo/static link flags.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-server/436)


bassosimone (approved):
lgtm

